PR #2, opened 2015-10-08
Update Gogs revision and state

mbugeia:


Kloadut:
Awesome! Do not hesitate to ask for an official integration if you are confident on the functionning state

Kloadut:
This application is pure genius.

And very very good packaging, you should ask for an official integration if you are willing to maintain it :-)

mbugeia:
Thanks, I may be a little too soon to an official integration, it seems that the install sometime fail (see https://github.com/mbugeia/gogs_ynh/issues/11). I prefer to wait for more feedback.

stevenroose:
Pure genius, it is!

PR #5, opened 2015-10-09
Helper script

Psycojoker:
Because I'm lazy as feuk.


opi:
thx a lot, lazy bstrd

PR #10, opened 2015-10-20
Make kanboard an official app

mbugeia:
Hi,
I'd like to propose Kanboard as an official Yunohost app.

Pros:
 - There isn't any official app that do project management
 - Kanboard is stable and in active development
 - Good integration with SSOwat
 - Package is easy to maintain/update

Cons:
 - UI is not very sexy
 - A little security bug that I hope will be fixed soon https://github.com/fguillot/kanboard/issues/1391

### DIFF
--- a/community.json
+++ b/community.json
@@ -245,12 +245,6 @@
         "state": "inprogress",
         "url": "https://github.com/julienmalik/jitsi_ynh"
     },
-    "kanboard": {
-        "branch": "master",
-        "revision": "161eebeaabf0f09035b7f68146be74eecf2ceeca",
-        "state": " ready",
-        "url": "https://github.com/mbugeia/kanboard_ynh"
-    },
     "kiwiirc": {
         "branch": "master",
         "revision": "c0aba5c4e2232d837299fe0cba12a962dd0f3bfa",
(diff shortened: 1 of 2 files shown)


julienmalik:
What are the other alternatives for kanboard-like apps out there ?

mbugeia:
Well there is some here:
https://github.com/Kickball/awesome-selfhosted#task-managementto-do-lists
and here:
https://github.com/Kickball/awesome-selfhosted#project-management

Most to-do list app are very simple and are more Trello-like than Kanboard-like, https://github.com/wekan/wekan (ex Libreboard) is a very good trello-like for example. But it doesn't even have a documented installation instead of docker/sandstorm/heroku and required Mongo DB wich is kind of heavy.

https://taiga.io/ is very good with a  beautiful UI but focused almost exclusively on Scrum, Kanboard is more flexible. And it doesn't have any kind of external authentification support.
https://www.openproject.org not tested but it required ruby and 512Mb RAM minimum

Also Kanboard is used for https://framaboard.org/

Kloadut:
Looks Good To Me :+1:

julienmalik:
+1

PR #19, opened 2015-12-02
update rainloop

Djip007:
bug fix and improvement:
- sso working,
- auto completion of mail suggest from ldap working
- domain autoconfig
- define admin password on install

### DIFF
--- a/community.json
+++ b/community.json
@@ -487,9 +487,9 @@
     },
     "rainloop": {
         "branch": "master",
-        "revision": "93c1f74fe309487406f65fde079ab3f09eaa0d10",
-        "state": "ready",
-        "url": "https://github.com/polytan02/rainloop_ynh"
+        "revision": "ecd879e2dc17fdeae029a543e92547a52b5a5a8e",
+        "state": "working",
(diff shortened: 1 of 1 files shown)


M5oul:
Thanks for that great work.

`working` status doesn't exist. You should use `ready`.

Djip007:
I use the scripte:
> python ./add_or_update.py community.json [github url]
... may be an error in the scripte 
thanks for the correction

Djip007:
Juste a question: the repos have change will the page https://yunohost.org/#/apps_in_progress_fr be auto-update ?

M5oul:
Ok.

app_in_progress page will be update. Actually, It's update manually by one person.

Djip007:
do you mean:
may be we should keep 'working' status and use 'ready' status were to make them official.
???

M5oul:
Conceptually, it was in this goal. This doesn't means, community apps can't become official.

Djip007:
OK!
Thanks for the "one" that make app_in_progress page update  :+1:

PR #22, opened 2015-12-28
Mark movim as ready

src386:
This time it's SSO ready.


jeromelebleu:
Nice, thanks! :) Is the HTTP Authentication now supported by Movim so?

src386:
yes, autologin with ssowat :)

PR #26, opened 2016-01-24
Update kanboard

mbugeia:


M5oul:
We need some test by other person before merging?

mbugeia:
It works for me but it would be better since I've done some change like adding a plugin that connect to LDAP to retrieve user infos (like full name).

mbugeia:
There is at least one user who have done the update with success https://github.com/mbugeia/kanboard_ynh/issues/21

PR #32, opened 2016-03-01
Add Keeweb

scith:
Keeweb is a password manager compatible with KeePass


M5oul:
Thanks for this package.

PR #33, opened 2016-03-01
Update community.json

tuxalex:
add some apps in development finded on yunohost forum

### DIFF
--- a/community.json
+++ b/community.json
@@ -65,6 +65,12 @@
         "state": "working",
         "url": "https://github.com/lunarok/cops_ynh"
     },
+    "coquelicot":{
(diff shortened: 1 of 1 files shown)


M5oul:
Sorry, I won't merge this pull request because:
- [Jeemdo have been removed because it's no more maintained](https://github.com/YunoHost/apps/commit/5d3c85e5310e3e36772f9732db17f945ca4594a9).
- My app coquelicot is not working at all
- Sonerez is an old package which I think need to be update.
- I don't know about zarafa.

tuxalex:
Ok, thanks for this reply.

PR #36, opened 2016-03-11
[doc] improve documentation regarding making an app official

Psycojoker:
Based of dicussions I've add in the past with Kload regarding the criteras.


Psycojoker:
(it would be great to be able to include that into Moul email to maintainers)

PR #43, opened 2016-03-31
[enh] don't reprocess github app if the revision is the same

Psycojoker:
This reduce considerably the time needed to build the result file and the number of hits on github api (therefor we don't need credentials anymore).


opi:
didn't review; just merge

PR #54, opened 2016-05-01
Move opensondage_ynh to YunoHost-Apps Github organisation

zamentur:


M5oul:
Ever done https://github.com/YunoHost/apps/blob/master/official.json#L50-L54.

PR #61, opened 2016-05-31
Update opensondage to 2.4

zamentur:


jeromelebleu:
Nice, thanks! Before merging, I think there is @opi comment to consider - regarding `nginx.conf` and PHP-FPM pool...

It's pleasant to see that helpers are used! :) I've just added some optional comments to use them just more... :$

zamentur:
Thanks for your comments!

I implemented it with f39a920ae9e5b9bf22dda1eb4e17d01dc7a64e52, I think
you could merge.

opi:
perfect !

PR #68, opened 2016-06-24
gnusocial app in progress state

gustavomrfz:
Still in development. 

Working, but with no LDAP and SSO so far.


M5oul:
It works for me.

PR #69, opened 2016-06-25
Add Apticron app

ldidry:


M5oul:
Thanks for the package!
As it's host on GitLab, it won't be possible to install it from web-admin which can only manage GitHub repositories.

PR #75, opened 2016-07-23
[up] official: transmission add back download button.

M5oul:
@opi,

You removed, then add back download button and other changes.
There is also other modification I don't what are they.
You didn't add it back on the list. Did you forget it?
Could this be merge?

### DIFF
--- a/official.json
+++ b/official.json
@@ -91,7 +91,7 @@
     },
     "transmission": {
         "branch": "master",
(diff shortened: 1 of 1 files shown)


M5oul:
Done with https://github.com/YunoHost/apps/commit/381e263e0d32a98c44e3b37d3d6e6cb17172d4bb.

PR #83, opened 2016-10-16
Make gogs an official yunohost app

mbugeia:
I'd love to make gogs an official yunohost app. I've updated the package to fix all the opened issues and to use new yunohost helpers.

Everything works well on my yunohost (including backup/restore) but the package need some external testing (and on other hardware, I only have x64 hardware).

Please let me know the result of your test and if some thing have to be improved :smiley:

### DIFF
--- a/community.json
+++ b/community.json
@@ -247,7 +247,7 @@
     },
     "gogs": {
         "branch": "master",
-        "revision": "0d23145c917ff168743c2ec071d865400350c076",
+        "revision": "46fb1398c7bad7718988c98572111c5fd25cbc9f",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/gogs_ynh"
     },
--- a/community.json
+++ b/community.json
@@ -245,12 +245,6 @@
         "state": "inprogress",
         "url": "https://github.com/YunoHost-Apps/gnusocial_ynh"
     },
-    "gogs": {
-        "branch": "master",
-        "revision": "46fb1398c7bad7718988c98572111c5fd25cbc9f",
-        "state": "working",
-        "url": "https://github.com/YunoHost-Apps/gogs_ynh"
-    },
     "hotspot": {
         "branch": "master",
         "revision": "850f19584efe0764642587bb49ba47e73415be6b",
--- a/official.json
+++ b/official.json
@@ -109,7 +109,7 @@
     },
     "wordpress": {
         "branch": "master",
-        "revision": "fe06981818b36b98256f56712da06bde301b73ce",
+        "revision": "034fcde60d7b512c0ded33f75e41a798b931100a",
         "state": "validated",
         "url": "https://github.com/YunoHost-Apps/wordpress_ynh"
     },
--- a/official.json
+++ b/official.json
@@ -17,6 +17,12 @@
         "state": "validated",
         "url": "https://github.com/YunoHost-Apps/dokuwiki_ynh"
     },
+    "gogs": {
+        "branch": "master",
+        "revision": "46fb1398c7bad7718988c98572111c5fd25cbc9f",
+        "state": "validated",
+        "url": "https://github.com/YunoHost-Apps/gogs_ynh"
+    },
     "hextris": {
         "branch": "master",
         "revision": "88d1916eceb760d62a9a3ea7bc5c4c6240a5840e",


M5oul:
Checked with `package_linter` and it's :ok: 
Need to be checked with `package_checker`.
I am using a previous commit of this new package which works very fine.
I will need to test an upgrade of Gogs.

maniackcrudelis:
I makes a test with package_check.
It's running...

maniackcrudelis:
I've obtained some errors

- Package_linter prints this warning: "✘ You should specify 'free' or 'non-free' software package in the license field."
- A malformed path may cause a crash of install
- Restore script seems break nginx and the adress return the default page "Welcome to nginx on Debian"

Otherwise, all works good.
And upgrade script works fine, but on a same version of app.

M5oul:
You could take a look at in progress [YEP (YunoHost Enhancement Proposal)](https://github.com/YunoHost/doc/blob/1432c21f8e7df88bf2a3ac942877b367c3e7e8be/packaging_apps_guidelines_fr.md).

mbugeia:
Thanks @maniackcrudelis I was not able make package_check working myself.

Regarding the license field I've indicated MIT which is more precise than free IMO. Is there a consensus on this ? I see some offcial app indicate the real license and some other just "free".

Do you have an example of an app which check that the path is correct ? 

I don't see why the restore script can break nginx but I will re-test it.

M5oul:
>Regarding the license field I've indicated MIT which is more precise than free IMO. Is there a consensus on this ? I see some offcial app indicate the real license and some other just "free".

Yep, we decided (me and @julienmalik) that app is using many libs with so many licenses. We decided to simply use (`free` and `nonfree`). @jeromelebleu is not agree with this choice.
In fact, we could only display main license (and not libs' one).

If so, we could use [SPDX format](https://spdx.org/licenses), which is also used on NPM (node.js) packages.

maniackcrudelis:
To check the path, I have a function:
https://github.com/YunoHost-Apps/leed_ynh/blob/master/scripts/.fonctions#L48-L55

For restore script, I search...

mbugeia:
First blocking issue https://github.com/YunoHost-Apps/gogs_ynh/issues/3 :(

maniackcrudelis:
I found the error in restore script.

https://github.com/YunoHost-Apps/gogs_ynh/blob/master/scripts/restore#L20

You can not use `sudo yunohost app checkurl` because the restoration has restored the ynh app conf. So, this path is already used by the app that will be restored

And, for an unknow raison, ynh_die seems print directly into the log without date and type of entry (ERROR or WARNING)

mbugeia:
Well for the restore I have a different opinion, I think package_checker does not test the correct thing since it backup the yunohost hooks and not only the app.

The backup restore process works well if you do:
```
yunohost backup create --ignore-hooks --apps gogs -n testbackupgogs
yunohost app remove gogs
yunohost backup restore --force --ignore-hooks testbackupgogs
```

It seems logical for me to test is the path is available before installing an app, you don't want to restore on a path not available. In fact most officials app have this check in their restore script.
But maybe I'm wrong, can someone on the yunohost team can rule on the question ? @jeromelebleu  maybe ?

jeromelebleu:
> You can not use `sudo yunohost app checkurl` because the restoration has restored the ynh app conf. So, this path is already used by the app that will be restored

In fact you can absolutely use this command as the `--app/-a` argument is given and contain the same app id - see [here](https://github.com/YunoHost/yunohost/blob/unstable/src/yunohost/app.py#L918-L921): it will check each app which defines this URL but the `app` one.

@mbugeia I agree with you, I do the same checks regarding app backup and restoration.

maniackcrudelis:
After many tests, it is not simply a problem with checkurl, but the use of checkurl on an app installed on the root of domain.
With or without hooks, checkurl crash if the app has restored is at the root.

This problem occured with package_check and on CLI.
But, actually, package_check tests backup and restore on root install by default. I will upgrade this test.

mbugeia:
@maniackcrudelis  You're right, that's indeed a bug !

maniackcrudelis:
Les nouvelles règles de validation étant validée, nous pouvons reprendre les processus d'intégration d'app officielles.
Les détails des niveaux sont ici:
https://github.com/YunoHost/doc/blob/packaging_apps_levels/packaging_apps_levels_fr.md

Le dernier test de gogs sur jenkins indique que le test de package linter ne passe pas.
https://jenkins-ci-arm.noho.st/jenkins/job/gogs%20(Community)/1/console

L'application semble utiliser ldap et j'imagine que la connexion via HTTP AUTH passe? Le cas échéant ton application peux prétendre au niveau 4.
Il faut au minimum fixer les erreurs avec package linter.

M5oul:
From my point of view, Gogs apps still need works to pretend official integration.
There is five issues with potential huge package modifications (gitea, move source path…).

Psycojoker:
Is it ok if I close this PR since we have changed the way we integrate official apps application?

According to the CI Gogs is level 3 for now so it needs a bit more work (but that still would be great to have it has an official application).

Josue-T:
For me it's ok. If nobody else work on this app it might improve it later when I have time.

JimboJoe:
OK for me.

Psycojoker:
Thanks for your answers :)

PR #86, opened 2016-10-28
[mod] ffsync: set app notworking and update hash.

M5oul:
I consider this is a minor vote, following rules needs to be completed before closing this PR:
- needs three votes. Two others. As mine is given doing this pull request.
- 66% of positives votes.
- vote will ends in one week: 4th November 2016.
- If you consider one this rules is wrong, please ask for modification.

Reasons:
- ffsync did not have commits since 6 months ago.
- It's not enough documented to make it works.
- [Many major install and sync bugs tickets](https://github.com/abeudin/ffsync_ynh/issues)
- Many user can't make it works (including myself).
- [Recent user can't make it works on the forum](https://forum.yunohost.org/t/firefox-sync-comment-que-cela-marche/2047)

If it works fine on your side, we could put it to `inprogress` status.

I ping contributors which could have make it works.
@abeudin, @julienmalik, @Josue-T, @lmangani, @matlink

### DIFF
--- a/community.json
+++ b/community.json
@@ -169,8 +169,8 @@
     },
     "ffsync": {
         "branch": "master",
-        "revision": "6eed9ad333d7411bf59b63f09562bc6513500507",
-        "state": "working",
+        "revision": "6707f664bbb39e18dd1ab6801d4c4524294e6c91",
+        "state": "notworking",
         "url": "https://github.com/abeudin/ffsync_ynh"
     },
     "filebin": {


maniackcrudelis:
Je suis pour

Il ne faut pas garder des app qui ne fonctionne pas. Autant que ce soit indiqué clairement. Elle pourra revenir en in progress si elle est reprise.

Josue-T:
I think mybe there are not a lot of work to do a app how work well, mybe we need to update this doc https://yunohost.org/#/app_ffsync_fr (it's done by this : https://github.com/YunoHost/doc/pull/406 and this https://github.com/YunoHost/doc/pull/407 ). It's not fully documented but just up to date...
And with the new update hash it could run because I had also an error 502 without this commit : https://github.com/abeudin/ffsync_ynh/commit/7df5d04f0c084a4bacf3139bb8636f0685bac5cf

But the other things is that actually I don't have time to fix all issues.
It's just my point of view... :-) I understand that it's not a really good app for all people.

M5oul:
Sorry, mistake from myself. Clic on wrong PR button :)
Pull request migrate to https://github.com/YunoHost/apps/pull/87.

PR #88, opened 2016-11-23
Fix redirect app commit revision

JimboJoe:
Actual commit id leads to an error when trying to install the app


opi:
yup, because we use https://github.com/opi/redirect_ynh instead of https://github.com/scith/redirect_ynh ; thanks for patching

PR #129, opened 2017-01-19
[mod] remove Let's Encrypt app from community list as YnH 2.5 now inc…

M5oul:
…lude LE.

### DIFF
--- a/community.json
+++ b/community.json
@@ -353,12 +353,6 @@
         "state": "working",
         "url": "https://github.com/Yunohost-Apps/lektor_ynh"
     },
-    "letsencrypt": {
-        "branch": "master",
-        "revision": "e498e05ebfc52880992507988571d24c35a3cd4c",
-        "state": "inprogress",
-        "url": "https://github.com/alexAubin/letsencrypt_ynh"
-    },
     "limesurvey": {
         "branch": "master",
         "revision": "c927cce4e76c4bd935e71fca2c11d03524bea332",


zamentur:
May be we should keep it to let people update their letsencrypt_ynh app before to removing it ?

M5oul:
Why do people needs to update LE_ynh?
Furthermore, `community` list must be added on the YunoHost instance to have an automated update.

zamentur:
Because without update LE breaks nginx when you remove it.

Psycojoker:
Should we prevent it from being installed in it's install script then? Once the new release has been made.

alexAubin:
> Because without update LE breaks nginx when you remove it.

It doesn't solve every problem though, AFAIK ;). But still important for people who installed it a long time ago I guess...

> Should we prevent it from being installed in it's install script then? Once the new release has been made.

Yep

M5oul:
I wan't to merge this one now. May we do it?

The job could have been done on `yunohost` repository to check time installation of LE_ynh (on setting.yml) and if it was install long time ago, propose to upgrade the app, remove it and then install LE with integrated certmanager.

Psycojoker:
iirc @alexAubin wanted to disable the app installation and add an upgrade part with a warning message before doing this.

alexAubin:
I made a commit and edited the README of letsencrypt_ynh do discourage people to install the app.

I don't know much about the implications of removing it from the json list, but if it doesn't break anything then okay :slightly_smiling_face:

frju365:
Perhaps you can ask for deleting the app repository too.

Psycojoker:
> I don't know much about the implications of removing it from the json list, but if it doesn't break anything then okay ￼:)

None that I can think of right now, the super small percentage of users that have added the community list as their list of lists won't received updates and that's all.

PR #147, opened 2017-02-12
change level of Garradin and scrumblr

frju365:


frju365:
It will be changed Friday...

PR #151, opened 2017-02-17
Update community.json

Josue-T:
Add riot and synapse and update monitorix and seafile


maniackcrudelis:
Do you really want to keep seafile as a notworking app?

Josue-T:
No, I think, you can put it as working.

PR #163, opened 2017-02-28
Upates glowingbear revision

jodeko:


frju365:
add after ""revision": ...,":
"level": 3,

community.json
```diff
@@ -269,6 +269,7 @@
     },
     "glowingbear": {
         "branch": "master",
+        "level": 3,
```
frju365: you can change to level 7 if you want now.

PR #169, opened 2017-03-04
Update community.json

rigelk:
- Replace Firefox Sync (ffsync_ynh) with an updated version (and also replace abeudin's apparently unmaintained repo).

See [discussion about my PR](https://github.com/abeudin/ffsync_ynh/pull/22#issuecomment-281053261) on abeudin's repo for more information about the proposed change of repository.

PR #177, opened 2017-03-05
Update community.json

rigelk:
- Replaces Firefox Sync (ffsync_ynh) with the now transfered repo to YunoHost-Apps.

Since I transfered ownership to YunoHost-Apps, the remote had to be changed as well here after #169 and [this bug](https://github.com/abeudin/ffsync_ynh/issues/23).


frju365:
thanks !

PR #183, opened 2017-03-11
Update official apps to catch up with "doc" commits

JimboJoe:
[YunoDash](https://dash.yunohost.org/index.php) is currently impossible to use (almost every application is red) due to many "organizational" commits (check_process, issues location, etc.).
In order to make things cleaner, this commit updates published revision to last updated revision for applications for which the newer commits don't include any software change.

[Minor decision](https://github.com/YunoHost/project-organization/blob/master/yunohost_project_organization_fr.md#d%C3%A9cision-mineure)
Will be closed on March 18th, or March 14th if decision anticipated.

### DIFF
--- a/official.json
+++ b/official.json
@@ -23,7 +23,7 @@
     "hextris": {
         "branch": "master",
         "level": 3,
-        "revision": "88d1916eceb760d62a9a3ea7bc5c4c6240a5840e",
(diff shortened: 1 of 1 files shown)


JimboJoe:
We reached the anticipated milestone with enough reviews, so this PR is merged.

PR #215, opened 2017-04-17
OS.js and Cryptpad package

frju365:


community.json
```diff
@@ -124,6 +124,12 @@
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/couchpotato_ynh"
     },
+    "Cryptpad": {
```
JimboJoe: The name doesn't respect [YEP 1.1](https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines_fr.md#yep-11---nommer-son-app-et-son-d%C3%A9p%C3%B4t---valid%C3%A9--manuel--notworking-) on package naming.

community.json
```diff
@@ -638,6 +644,12 @@
         "state": "working",
         "url": "https://github.com/lunarok/openwrt_ynh"
     },
+    "OS.js": {
```
JimboJoe: Same here.

frju365:
it's ok for Cryptpad but it's not for OS.js (original name of the app... so I close the issue)

PR #232, opened 2017-05-16
Deprecate Jappix app

M5oul:
Jappix is no more maintained since a while.
Moved to community list.
There will be no more web XMPP client on official list.
This must encourage us to push Movim app to official list.

### DIFF
--- a/community.json
+++ b/community.json
@@ -383,6 +383,12 @@
         "state": "working",
         "url": "https://github.com/JocelynDelalande/ihatemoney_ynh"
     },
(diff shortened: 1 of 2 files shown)


scith:
Ok

maniackcrudelis:
I'm ok with that, if it's followed by an announce to clearly inform the potentials users.

M5oul:
- [Jappix app is deprecated](https://forum.yunohost.org/t/jappix-app-is-deprecated/2916)

PR #251, opened 2017-06-07
Make Garradin official

frju365:
Hello,
I wanted to add Garradin as an official package. I need more tester for that.
CI-Level = 7

I need test for : 
- [ ] Backup
- [ ] Restore
- [ ] Restore Database
- [ ] Upgrade

- [ ] I need a review of the code.

~~I need your point of view too : is it better to wait for another release of Garradin or we can make this official ?~~ https://github.com/YunoHost-Apps/garradin_ynh/pull/4

Thanks,
frju365


scith:
So did you manage to get SSO working?

scith:
And automated setup (with curl or something) ?
Let me know if I can help (although my curl apps are not passing pcheck...)

frju365:
For SSO, it doesn't work at all or we must create a extension/patch of the app. I asked the dev for the LDAP arch and he said that he will perhaps think of it but in the future, not now.

The automatic setup is good... but in a way I prefer letting the president of the association create his own config. The only way is to use curl... however, I don't think using curl is a good idea.

scith:
We could also install, import users abd log users in with custom PHP scripts (also called following post_user hooks).

The default install (like association name, address, mail and members categories) could be customized afterwards I think.

I know these are hacks, but I am not sure I am in favor of official apps that still need manual installation, manual creation of users and non-automatic login ... Especially when there are possibilities to avoid these

What do you think?

frju365:
As it's impossible to use SSO for now... I don't know. The last app I see using SSO is now broken (flarum) and I don't want to break my package. For the setup, I can do something, but not for the SSO login. ("Centralize" is good but we can chose what we centralize too and I've seen many time my LDAP table or my SSO -- permission per user -- broken without doing anything -- so now, I've many doubts about this "all apps must use SSO/LDAP".)

JimboJoe:
I tend to agree with @scith regarding official apps' quality: in my opinion YunoHost is mostly judged based on these applications functioning well, and being easy to install. And the SSO integration is a must-have in my opinion, as YunoHost aims at demonstrating seamless integration of every application.

As for your opinion @frju365, I'm not sure your unfortunate experiences can ben generalized: my LDAP never failed me, and the Flarum case must be analyzed. What it demonstrates is that getting LDAP support in application that don't offer it officially is a significant effort that the Apps Group can barely afford...

frju365:
ok... I agree with you ( @JimboJoe , @scith ). It's ok.  But I will surely need help for that. I will investigate the code and see if I can make something.

alexAubin:
Hm, what's the status of this ? It feels kinda weird to have this PR now that the standard process seems to make polls to know which app should become official ?

maniackcrudelis:
The new method with polls is only a way to add new official apps even if nobody ask for it.
But, I think we can (should ?) keep also this way if someone want to add his app and is ready to maintain it.

Psycojoker:
> I still don't know what I must work or not... becaues in fact, I had
> feedbacks from person who are quite against using the sso for an
> association, moreover if it's seprated from the main use of the
> server.

Yes, there is a discussion about "should we make activating the SSO an
option or not" for application in redmine somewhere (or at least the
discussion as started).

Those are 2 differents usecases and we get more and more people not
interested by having the SSO (I guess to use yunohost to do classical
deployments, not "for users in the SSO").

That's a log shot but I'm thinking about having either a new
"toggle-sso" live "change-url" or simply finally writting those damn
"actions" for the applications.

frju365:
In fact, if I want to have LDAP in Garradin, I have two possibility : 
 - Dev the app to create a php script and integrate it in the app after that. I don't even know if it will be secure enough.
 - Wait for the dev who will perhaps in a future include an LDAP feature.

frju365:
I don't know what we must do with this PR...

frju365:
It's either there are some persons who agree with merging this PR or we close this PR. :-) Because this PR was made more than 6 months ago.

alexAubin:
Well I don't have any strong opinion, but to help me understand the situation and by curiosity : 
- you mentioned in a comment previously that some people are against using the SSO for this kind of app : can you elaborate on the motivation behind this ?
- does this app allow the creation of accounts by anonymous users (e.g. people willing to register as member of the association, similarly to galette) ?

frju365:
For, SSO, I gave up. :-) This app doesn't integrate LDAP and I don't have time to implement it. Second, the association owner/president is perhaps not the owner of the server so, he doesn't specifically need SSO. Third : SSO and an association with 150 members are incompatible (we can't do search of user in the admin UI, so no...).

maniackcrudelis:
Sorry, but I just looked at this app, and it seems like an old pile of codes abandoned in a forgotten place of github.
I mean, I don't know this app, I don't use it, but just, to pretend to be official, at least you can maintain it.

My impression about this it's that this app is going to be an old crap we will have to maintain by ourself, hoping that we'll never have to open an issue on the upstream "repository" (!!!)
So, it's just my impression, I didn't try this app. But, just the look it has, it's not really encouraging...

frju365:
as, it's me who maintain it,... yes, I maintain it. Normally it the latest version. But I didn't change anything in the code, as a PR was opened but nobody answered. Sorry, but your point of view is true... but I waited to have a point of view on this PR. :-(

frju365:
As nobody answer during quite answer to this PR, I close it. Thanks @scith ,@JimboJoe, @alexAubin and @Psycojoker for your comments.

PR #262, opened 2017-06-20
Update ihatemoney version and level

JocelynDelalande:
For the level, I think it meets the requirements for llvl 7 **except for multi-instance installation** (that I didn't implemented, it has little interest IMHO, as ihatemoney is though as a public service). 

But I'd like your opinion on that ? should I put lvl 7 or 2 till multi-instance is supported ?

### DIFF
--- a/community.json
+++ b/community.json
@@ -384,8 +384,8 @@
     },
     "ihatemoney": {
         "branch": "master",
-        "level": 1,
-        "revision": "309e46013fdf6909f741dc78845ae6115d1f2071",
+        "level": 7,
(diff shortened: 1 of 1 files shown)


JocelynDelalande:
@JimboJoe ok, thanks for the merge, but that does not answer my question ;-)

JimboJoe:
Sorry I missed it!
Technically you didn't need to set the package level: it is automatically assessed and updated every Friday.

On your precise point on multi-instance, provided it's usually quite easy to implement (provided you use $app everywhere possible), it is a good practice to implement it.
Don't hesitate to keep asking for help in the XMPP #apps channel ;-)

PR #267, opened 2017-06-23
Update app's level

maniackcrudelis:
We're waiting for nextcloud, zeronet and limesurvey.
Those previous tests are false.

But, the CI is busy with a lot a testing and unstable tests...


maniackcrudelis:
Can be merged now.

PR #288, opened 2017-07-16
Added working version Yourls with latest update

anmol26s:
1.Changed the nginx.conf file so that the curl works.
2.Deleted sources and added app.src in conf so that md5 checksum can be checked
3.Changed the install script so that the issue with double trailing slash could be resolved see https://github.com/courgette/yourls_ynh/issues/4
4.Updated the remove script with latest YunoHost requirements.
5.Added update,backup,restore,check_process and license.
6. Finally updated Yourls to version1.7.2.

### DIFF
--- a/community.json
+++ b/community.json
@@ -1037,9 +1037,9 @@
     "yourls": {
         "branch": "master",
         "level": 1,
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Thanks for this huge work !

PR #326, opened 2017-09-03
phpldapadmin upgraded and working

aymhce:


maniackcrudelis:
https://github.com/YunoHost-Apps/phpldapadmin_ynh/issues/2#issuecomment-326834189

maniackcrudelis:
Currently, this app failed to these tests:
- Upgrade
- Multi-instance installations
- Restore

So in state, this app can't be level 6. Only level 1 if you want to upgrade your app like that.

aymhce:
Ok thanks for report. I have disable multi instance test, there isn't this function. I do a fix for restore. But I don't known with Upgrade don't work. I have no pb with it at home.

However, can you retest please since last commit please ?

maniackcrudelis:
When you do an upgrade, the nginx config file has a blank location
```
location  {
```
That's why nginx fail to reload.

aymhce:
Ok ! It's in case on root path. It's ix on last commit :)

PR #329, opened 2017-09-08
Update app's level

JimboJoe:
CI has been re-launched for ampache and monica apps, waiting for results.


JimboJoe:
ampache raised to level 3.
monica still fails, [issue created](https://github.com/YunoHost-Apps/monica_ynh/issues/8).

PR #340, opened 2017-09-25
Add new app Portainer

aymhce:
Hello,

For me, it's working, but this app is inside a docker image.
I don't know if Yunohost CI (on LXC) is docker compatible... Can you test it please ?

Thanks

### DIFF
--- a/community.json
+++ b/community.json
@@ -918,11 +918,12 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/pluxml_ynh"
     },
-    "Portainer": {
+    "portainer": {
         "branch": "master",
         "revision": "a7e7e2faa5bc21bd2c2685c958375a05d9ccf70b",
         "state": "inprogress",
         "url": "https://github.com/aymhce/portainer_ynh"
+    },
     "prestashop": {
         "branch": "master",
         "revision": "34b6f2e2f33c528a411a5c1690ad7124c68bb6da",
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
It doesn't work for me.
`_common: ligne 20: DOCKER_HOST : variable sans liaison`

aymhce:
Ok I (think) fix it . It was in case you are with native yunohost. I have yunohost in docker. Can you retry please ?

maniackcrudelis:
Same error.
You can't use `[ -z $DOCKER_HOST ]` just like that because you've used before `set -eu`. So if $DOCKER_HOST is unset, your if will be broken immediately.

You have to set this value before or to use a OR to go through the `set -eu`

aymhce:
Thanks for help (I doesn't know set -eu ...). I think I've find a solution. Can you re-test please the last commit ?

maniackcrudelis:
Docker failed to install
```
curl -sSL https://get.docker.com
# Executing docker install script, commit: 490beaa
Either your platform is not easily detectable or is not supported by this
installer script.
```

aymhce:
Ok sad... What is your plateform ? Do you use lxc ?

maniackcrudelis:
All the tests are perform in an lxc container.

aymhce:
Ok, next week I'll work about docker integration in lxc for passing tests.

aymhce:
hello @maniackcrudelis after a long time ago, here is a new commit ! Can you test it please ?

maniackcrudelis:
It's all ok for me now ;)

PR #382, opened 2017-11-20
Update hotspot_ynh and vpnclient apps

alexAubin:
Today a user had an issue where he manually installed an Internet Cube, did not have the list from LaBriqueInternet, but did have the community list. He tried to install hotspot_ynh and got stuck with lock issues because the commit was ~1 year old (c.f. recent commits from the Brique Camp).

### DIFF
--- a/community.json
+++ b/community.json
@@ -882,7 +882,7 @@
     "piratebox": {
         "branch": "master",
         "level": 2,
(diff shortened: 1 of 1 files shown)


JimboJoe:
I also updated piratebox entry (from LaBriqueInternet as well).

PR #388, opened 2017-12-15
Modify level

maniackcrudelis:


JimboJoe:
How come jirafeau is at 0...?

maniackcrudelis:
The install script fail. That's why there's a testing right now.

PR #404, opened 2018-01-13
Upgrade Nextcloud to 12.0.4

JimboJoe:
I'm taking care of the forum thread.


maniackcrudelis:
Can we merge anyway ? Or you prefer to have a forum thread before ?

PR #407, opened 2018-01-20
Update app's level

maniackcrudelis:


JimboJoe:
@frju365 what about chticky notes at level 0? Did you re-launch the CI or contact the maintainer?

PR #414, opened 2018-01-31
Upgrade Mattermost

kemenaran:
Changes:

- Upgrade to Mattermost 4.6
- Working backup and restore
- Systemd service
- Integration into Yunohost status panel
- Multi-instance supported

PR #415, opened 2018-01-31
Revert "Upgrade Mattermost"

kemenaran:
Reverts YunoHost/apps#414

It seems the upgrade fails for some users – and in the failure process, Yunohost fails to restore the previous version of the app, which has to effect to remove the app from the server :/

See https://github.com/YunoHost-Apps/mattermost_ynh/issues/90#issuecomment-361880523

@JimboJoe sorry for the noise. Can I let you merge this while I investigate the issue?

### DIFF
--- a/community.json
+++ b/community.json
@@ -631,8 +631,8 @@
     },
     "mattermost": {
         "branch": "master",
-        "level": 7,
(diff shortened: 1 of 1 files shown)


JimboJoe:
No problem! 😉

PR #437, opened 2018-03-11
Add SPFToolbox

frju365:
SPFtoolbox is a nice app to check record and domain name settings. Perhaps one day hosted by Yunohost. :D 
https://github.com/charlesabarnes/SPFtoolbox
http://spftoolbox.com/

### DIFF
--- a/community.json
+++ b/community.json
@@ -1085,7 +1085,7 @@
     },
     "spftoolbox": {
         "branch": "master",
-        "revision": "9c5764997ff1f62c94e465a574f406b8e53eebfb",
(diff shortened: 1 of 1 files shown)


JimboJoe:
Looks promising!
The README states "Inprogress/NotWorking", so... is it working or not? :grinning:

frju365:
ah yes, I forgot to change it ;)

frju365:
It's ok now.

PS: Yes. I ask the dev to add an open/close port feature. After that, I think it will be a very complete app to check DNS/host/Records.

PR #461, opened 2018-04-02
Move collabora to notworking

Rafi594:


frju365:
Just inprogress perhaps instead of NotWorking ?

Rafi594:
This applictation does not work on Jessie but it works on Stretch

frju365:
Ok. Anyway, it's your choice ;) Let's merge it!

PR #486, opened 2018-05-05
[enh] Add app logos to the official list

Rafi594:
Not ready yet.


M5oul:
It would have been better to use SVG files as they are much more lightweight for a git repository.
I am sure you won't found a SVG file for all of them, but at least some.

zamentur:
What is the goal of this PR ? Be able to display logo in app list ?

Rafi594:
Yes for display apps logo here https://yunohost.org/#/apps

alexAubin:
So what do we do with this ... On my side I'm not convinced that this is a good idea because logos are always a mess to handle : they need to be standardized, you need to have *all of them* (and those are just officials) for it to make sense, and they always look different which imho does not make it prettier than simply displaying the text (or c.f. the look of SSOwat)

Also I don't think it makes any sense to add a new key where the info about the .png file can be deduced from the name of the app : `images/<appname>.png` ...

alexAubin:
Ok guys, closing this ... There's a lot more work and discussions to have about this if we really are going to do anything serious with app logos. Since it's been inactive for ~6 months, I'm closing. Feel free to reopen/complai if you want to discuss/work on this.

PR #510, opened 2018-05-23
Mumble standardization

Jibec:


Jibec:
let's wait one more minute

JimboJoe:
Some remorse... ? :thinking:

Jibec:
:D
This is old work, but I want to double check upgrade from past install to make sure it works fine.
I ran tests for debian 9 and noticed Check_process file is uncomplete.
It's better to fix this before having sad users. It shouldn't be too hard.

PR #511, opened 2018-05-24
Use HEAD for all official apps

alexAubin:
Alright guys, it's like the 4th time recently that we have troubles because we forget to upgrade the commit in the app list and this is already something that we raised during the Brique Camp

So I'm bringing it on the table : let's get rid of this burden of "oh yes we forgot to update the commit". The process of merging things is already quite complicated between the PR review, the testing branch and the delays for merging them. We can and should get rid of this easily-forgotten step. So let's use HEAD for all official apps ...

The only small known issue is that it might trigger an update in YunoHost even if only the README got changed for instance ... but this can also be improved later (looking at the diff and not updating the commit in the built list if only the README changed). Overall that should only be a negligible issue...

### DIFF
--- a/official.json
+++ b/official.json
@@ -2,35 +2,35 @@
     "agendav": {
         "branch": "master",
         "level": 0,
-        "revision": "0259831957d005c64772080c85e81b03604d760f",
+        "revision": "HEAD",
         "state": "validated",
         "url": "https://github.com/YunoHost-apps/agendav_ynh"
     },
     "ampache": {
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
If we're sure that we can trust that, let's do it.
Anyway, there isn't going to be any upgrade of readme without a testing. Since we're not merging in master.

maniackcrudelis:
Yolo, let's breaks official.json :D

PR #514, opened 2018-05-25
Update synapse

Josue-T:
Following https://forum.yunohost.org/t/synapse-install-bug/4905 and https://github.com/YunoHost-Apps/synapse_ynh/issues/49 I think it's the time to update the commit :-)


frju365:
Already set to HEAD in Master branche. Let's close this PR.

PR #517, opened 2018-05-26
[enh] Support gitlab

zamentur:
Adapt add_or_update.py to support gitlab


Psycojoker:
Since it's a minor script only periodically used I don't see the point of waiting much more

PR #518, opened 2018-05-26
[enh] Put HEAD if a testing branch exist

zamentur:
Put head in revision if a testing branch exist on the repository

Note I have built this on top of gitlab PR https://github.com/YunoHost/apps/pull/517

### DIFF
--- a/add_or_update.py
+++ b/add_or_update.py
@@ -4,7 +4,7 @@ import os
 import sys
 import json
 
-from urllib2 import urlopen
+from urllib2 import urlopen, URLError
 
 states = {
     1: "notworking",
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
If the idea is to put HEAD to all apps with a testing branch, I don't think that's a good idea.
Using HEAD instead of a commit should be a decision from the maintainer of an app, and also clearly be known by the maintainer himself.

Psycojoker:
Looks a bit way to automatic for me too, that could end up in unexpected situations and bad results :/

alexAubin:
That's when using `add_or_update` though, we're not talking about the list building part here

zamentur:
What about this PR ?
Is the argument of Aleks enough ? Or shall we close this PR ?

maniackcrudelis:
What would be the point of this PR ?
I mean other than adding automatically HEAD if there's a testing branch, for what we already said that would be a bad idea.

zamentur:
As explained, by Aleks, it put HEAD only for new apps with a testing branch added with add_or_update.py script, if you understood that and continue to think it's a bad idea, feel free to close the pr. I was just asking if you understood correctly the scope of this change.

So feel free to close this pr if for you it's problematic to put HEAD in this situation.

maniackcrudelis:
What's problematic for me is to put HEAD for an app without the explicit action of its maintainer.
We can't assume any maintainer, using or not this script, will know the consequences of using HEAD for an app.

alexAubin:
Alright guys I guess we can close this, last time we discussed it with ljf it was not really a huge deal...

PR #506, opened 2018-05-21
Notworking apps

maniackcrudelis:
A batch of notworking apps, this PR should wait until May, 27. Accordingly to the issues opened in each app repository.

https://github.com/YunoHost-Apps/chtickynotes_ynh/issues/6
https://github.com/YunoHost-Apps/cryptpad_ynh/issues/24
https://github.com/YunoHost-Apps/funkwhale_ynh/issues/7
https://github.com/YunoHost-Apps/gitea_ynh/issues/7
https://github.com/YunoHost-Apps/gitlab_ynh/issues/12
https://dev.yunohost.org/issues/1130
https://github.com/YunoHost-Apps/hubzilla_ynh/issues/24
https://github.com/YunoHost-Apps/ihatemoney_ynh/issues/23
https://github.com/YunoHost-Apps/kodi_ynh/issues/4
https://github.com/LowMem/kresus_ynh/issues/12
https://github.com/YunoHost-Apps/linuxdash_ynh/issues/5
https://github.com/YunoHost-Apps/monica_ynh/issues/8
https://github.com/YunoHost-Apps/nexusoss_ynh/issues/4
https://github.com/YunoHost-Apps/osjs_ynh/issues/10
https://github.com/YunoHost-Apps/owntracks_ynh/issues/7
https://github.com/YunoHost-Apps/pelican_ynh/issues/3
https://github.com/YunoHost-Apps/ulogger_ynh/issues/1
https://dev.yunohost.org/issues/1129
https://github.com/YunoHost-Apps/Webtrees_ynh/issues/3
https://github.com/YunoHost-Apps/yunofav_ynh/issues/6
https://github.com/aymhce/yunohost_ynh/issues/1

PR #525, opened 2018-05-28
Updated few apps

anmol26s:


alexAubin:
About webtrees though : the app was found to be level 0 and therefore flagged as notworking by https://github.com/YunoHost/apps/pull/506

I don't see any new commit on the apps to try to fix it ? :s

anmol26s:
I can not access the fail CI anymore. I use the app on my server. The core app has not been updated from quite some time, so didn't got a chance to touch the app from long time. But anyway I will test it again and see for CI errors.

PR #529, opened 2018-06-05
community.json: add gitlab-runner

kemenaran:
Add https://github.com/kemenaran/gitlab-runner_ynh . It enables users to run CI jobs for their GitLab projects.

### DIFF
--- a/community.json
+++ b/community.json
@@ -440,6 +440,13 @@
         "state": "inprogress",
         "url": "https://github.com/YunoHost-Apps/gitlab_ynh"
     },
+    "gitlab-runner": {
+        "branch": "master",
+        "level": 0,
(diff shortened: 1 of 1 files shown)


community.json
```diff
@@ -440,6 +440,13 @@
         "state": "inprogress",
         "url": "https://github.com/YunoHost-Apps/gitlab_ynh"
     },
+    "gitlab-runner": {
+        "branch": "master",
+        "level": 0,
```
alexAubin: Not sure you have to / should set the level here (it's usually set semi-automatically via the CI result) but that shouldnt hurt

JimboJoe: Yes, you can definitely skip that!
It will be set with the real value on next Friday anyway :wink:

PR #530, opened 2018-06-06
[Translation] add a space

Kcchouette:


Psycojoker:
Hello @Kcchouette,

Translation needs to be done on https://translate.yunohost.org otherwise they will be overwritten :/

Therefor we can't accept this proposition here, sorry (ping @Jibec)

Psycojoker:
Oh and thanks for your contribution :heart:

Jibec:
thanks you @Kcchouette , both corrections are now done in [gitlab](https://translate.yunohost.org/translate/yunohost-apps/apps/fr/?checksum=3a3262f1dab3d5f7) and [gitea](https://translate.yunohost.org/translate/yunohost-apps/apps/fr/?checksum=ba923ff69a0a96fe)

If you want to translate a specific package: use search feature on "Context" field, or download the csv file and upload it :)
Please ask me if any help is needed (https://mamot.fr/@jibec).

Kcchouette:
Hello, thank you all.

It's sad that this information (=that translation is done at https://translate.yunohost.org/)  is not written in the README

maniackcrudelis:
Wouldn't be more useful for all contributors to be able to say that there's a mistake somewhere in a package translation, and have some guys from translation team hanging around to answer that kind of things !?
I don't think that it's really efficient to say to everyone who see an error, go to translate.yunohost.org, try to find your way through this mess, and maybe, if you're brave enough (I wasn't, I give up before finding how to fix anything) you could eventually try to fix the error you've seen..

Jibec:
Well, it's not forbidden to file issues. What's the real problem you're facing?

maniackcrudelis:
I love to be taken for a asshole...

Things have been said anyway, users should be able to say if there's a mistake in a translation file in this repository. And be listened when they do so.

JimboJoe:
Hmm, @Jibec has promptly:
- replied stating that he had applied the PR recommended changes via weblate
- apologized for this submission process not being transparently described an proposed his help
- and gently asked for where that could be documented.

Afterwards you ranted on weblate being a "mess", and interpreted @Jibec's last answer as ironic whereas I think it's pragmatic.

In both cases, I think this is not the appropriate forum. There seems to be a conflict between you two. It didn't bother me at first, but now it's beginning... Could you please be kind to the project and deal with your conflict once and for all? Thanks :innocent:

Jibec:
Well don't take it that way, do you mean you want to allow PR for translation or enable issues ?

Inviting to use translate.yunohost.org is for multiple reasons :
1. Advise what we can translate and consolidate our translator base
2. Bring translators a way to get notification on changes and comments
3. Lower the technical skill required to contribute
4. Make sure translation discussion are handle between translators

Here we explained the way to contribute and I did the requested changes.

I think we should continue to invite translators to use the dedicated platform.

maniackcrudelis:
> There seems to be a conflict between you two.

I was expected that since a long time...
But actually, no, there's nothing personal. Even if there's many subjects I disagree with Jibec (since December 2016...)
I have conflict with concepts:
- Not respecting what other people have written and modifying it without asking them. (Problem solved by a commit from Bram preventing string replacement. I'd rather prefer a concrete pre-translation solution.)
- Link to the previous one, ultra-francization, which have been too much time debated in our xmpp room but nevertheless applied many time against my will.
- Globally, translation team (not only Jibec...) who overwrite and translate strings written by packagers without their consent or their approbation, not even a look at it. There's probably many packagers who don't know that the files in locales/ will be used instead of their original manifest.
It's highly disrespectful for the work of those packagers. That's why I disagree with this method since the beginning.

Still there's nothing personal, I rejected a PR from Jibec on Package check for good (and widely explained) reasons. As I did before with others, especially Fr0g with his beginning of refactoring.
I rejected other PR also for good reasons, as explained each time...
We had other conflicts on the forum on other matters, and still there's nothing personal.
As I said before, I'm not nice, I'm not kind, I'm not politically correct, and I never tried to be so. But, there's one difference with Jibec comparing to others, Jibec is the only one, so far, to have cross the limit of my patience. So that I'm even less nice and kind. That's what brings to cross my limits, I have less filters in what I said. Hopefully, I have a lot of filters...

And the main reason why he has crossed my limits is, as I already said loudly on the xmpp room about one year ago (though I can't find it in my archives), that he's still sneaking in my back to push his ideas even if I'm against, and continually playing Calimero and vanishing instead of confronting his idea.
And that's personal !

Finally,
> Could you please be kind to the project and deal with your conflict once and for all?

There's no conflict to deal with, I'm just trying to do what I have to do, which is also defending the work of all current packagers (even Jibec himself), previous ones and those to come. On every matters I have to.

Jibec:
we indeed had multiple opportunity to disagree, and I learned a lot from you, said thank you and told you when I was wrong. Latest public example was searx.

You not being personal doesn't imply it doesn't hurt me.
Being connected a lot and having an important role in the community shouldn't allow to increase our bad side. Quite the opposite.

Your point of view is valuable and whatever how harsh it was, it made me progress, only the way to say it is hard to endure over time.
I have very few spare time, I don't bother answering when solicitation is not direct. The way you reacted to the postgres helpers was just too much for me to handle, following multiple other useless indirect fights we had over time.

But I understand most of the issue is translation related here, which is an interesting debate, let's have it.

We are french, and are lucky because we have multiple open-source contributors able to translate in our language. So we have the "critical size" for dev contributors to keep the control over the language, and reviewing the way we should speak.

Most developer doesn't bother using English world all the time because they are educated with English usage and mix up words all the time. This doesn't mean the end user should endure English words in French interface and try to figure out what it may mean.

In the same time, most languages have a really little contributor base. Using Pull Request make it hard to understand where to contribute. We would never have such a high level of Arabic and Occitan we that rule. With Arabic, we now have a better example to show how the maintainer can't challenge the translation. You just have trust and look for a way to welcome peer review and translator discussions. And that's the normal translator-dev relation we should target.

I do agree translate "email" to "courriel" is political, and I assume doing it, I'm not alone, most translators do it. They still and will always be people to nitpick about it. That's life.
In the same time, I don't understand how the word "package" should not be translated as even in the rare mumble discussion I attended, I heard multiple time the translation used.

Remember I often say this: 100% translated is not enough, context is really important for translation and we often do mistake. My English is quite bad, even my French, and it's ok to correct it.

What Bram did here of package translation is a great improvement for non English speakers.
It's the same work [Debian](https://www.debian.org/international/l10n/ddtp) and [f-droid](https://hosted.weblate.org/projects/f-droid/fdroiddata/) do, and I don't think it doesn't respect maintainers work.

What we did here regarding to maintainers' work is worse: https://github.com/YunoHost/yunohost/pull/476 and it still is a great improvement for end-users (english native or not).

That's why you should help to remove the opportunity to write another language than English in manifest.
And accept weblate's contribution can override manifest translations (which you blocked).

PR #550, opened 2018-06-21
add pixelfed

Jibec:
Answers https://github.com/YunoHost-Apps/pixelfed_ynh/issues/1

fixes https://github.com/YunoHost-Apps/pixelfed_ynh/issues/1

### DIFF
--- a/community.json
+++ b/community.json
@@ -54,6 +54,12 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/pagure_ynh"
     },
+    "pixelfed": {
+        "branch": "master",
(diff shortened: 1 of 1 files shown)


Jibec:
@JimboJoe if we should not use capital letter, then I suggest to correct the whole file.

frju365:
bon, c'est pas très très grave comme erreur. Normalement, on ne doit pas utiliser de majuscule mais bon, c'est ok si il y a une raison valable.

frju365:
Moreover, it's alphabetical order, so it's after Pelican and not before.

frju365:
Done.

PR #570, opened 2018-07-26
Add Discourse

JimboJoe:
Now that the CI seems stabilized, let's see how the CI handles that! :innocent:

### DIFF
--- a/community.json
+++ b/community.json
@@ -238,6 +238,12 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/diaspora_ynh"
     },
+    "discourse": {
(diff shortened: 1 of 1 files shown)


JimboJoe:
Your PR got merged, thanks @tituspijean !
I've upgraded to upstream version 2.0.3, which was tough because the ARM compatibility was broken (upstream doesn't take care of ARM...).
Now in latest tests, installation fails because a web server of some dependencies (xmlsoft.org) is down...
Let's hold on a little bit, then... :sleeping:

JimboJoe:
Everything looks good to me now!

PR #576, opened 2018-08-16
Set sogo as maintened and change url

Josue-T:
fixes https://github.com/mattronix/sogo_ynh/issues/3

### DIFF
--- a/community.json
+++ b/community.json
@@ -1250,10 +1250,10 @@
     },
     "sogo": {
         "branch": "master",
-        "maintained": false,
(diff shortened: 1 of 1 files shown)


frju365:
If it's maintained, should we add this line, or can we just delete it ?

Josue-T:
I don't know, many apps don't have this line.

frju365:
Well, I'm for remove it, as it's not fully implemented I think.

alexAubin:
You can either delete it or set it as true :+1:

frju365:
ok. If it's the same it's ok

PR #598, opened 2018-10-14
Update app's level

maniackcrudelis:


maniackcrudelis:
As reminder, don't forget that those PR have to be merged within a week, otherwise no other commit can be make by the script.
So those, very easy to review, PR should be quickly merged, and then remove the branch to allow the following PR the next Friday.

PR #623, opened 2018-11-24
Update app's level

maniackcrudelis:


maniackcrudelis:
Meanwhile, those apps aren't working in our CI.
As well for any user using LXC.

So we can't give any level to those apps. Either they're inprogress or off list, either they're level 0...

frju365:
ok. No problem. I do it in the next days (today or tomorrow)

alexAubin:
:| ...

So what do we do with this...

Personally I think level 0 is fine, from the point of view of the users, it's not worse than not having levels at all. It would be nice to have a solution to test those apps automatically but doing manual tests is just a waste of time and in the meantime that's not energy spent in automatizing tests for dockerized apps.

Can't we just move on with this semi-automatic PR ? :|

maniackcrudelis:
Anyway, if it's not today, it will be tomorrow, those apps are going to be set at 0 by the next round of the script.
So, meanwhile, let's merge to not block the update of the other apps.

In the check_process you can also fix the level at 1 or whatever, or find a real solution to check those apps.

maniackcrudelis:
Well...
It's time I guess for my autocratic merge :smiling_imp:

PR #627, opened 2018-11-29
[fix] Burry some package with no plan to be maintained

zamentur:
Here i suggest to remove some apps from the community list.
I have not removed jappix cause the app is working. We probably miss a way to indicate an app is deprecated.

### DIFF
--- a/community.json
+++ b/community.json
@@ -45,14 +45,6 @@
         "state": "inprogress",
         "url": "https://github.com/YunoHost-Apps/agora_ynh"
     },
-    "ajaxgoogleapis": {
-        "branch": "master",
-        "level": 0,
-        "maintained": false,
-        "revision": "17565a0ed6c6f617a53d36f5b91c1a7ea884a11a",
-        "state": "notworking",
-        "url": "https://github.com/zamentur/ajaxgoogleapis_ynh"
-    },
     "apticron": {
         "branch": "master",
         "maintained": false,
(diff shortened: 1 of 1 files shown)


JimboJoe:
I'm OK with burying applications for which the upstream is definitely dead (like shout which has been forked and abandoned).
I think we shouldn't burry applications if that's not the case: for instance, owncloud is still maintained and "may" someday be more attractive than it is today versus Nextcloud. And that day, we want the volunteer owncloud app packager to know this package existed... Don't we? :smile:

maniackcrudelis:
Agree with JimboJoe and frju365

zamentur:
etherpadlite_ynh has not been updated since 4 years... The package etherpadmypad seems replace it.
About owncloud it's like you want.

alexAubin:
Bump :s

JimboJoe:
@frju365 in particular :wink:

frju365:
For etherpad lite, I'm still against. But, as we have etherpad_mypads which have better features, it's ok.

frju365:
@maniackcrudelis : Can etherpad_mypads replace totally etherpad_lite ? (or do we must maintain it too ?)

alexAubin:
Bump ... what do we do with this :s

JimboJoe:
etherpad-lite looks [still maintained](https://github.com/ether/etherpad-lite), so my opinion about the corresponding package is the same as for owncloud: keep the package in the list, if someone wants to maintain it.

maniackcrudelis:
Actually, frju365 is right about etherpad-lite. etherpad_mypads can be considered as a maintained replacement package of etherpad_lite.
During the installation, you can choose to not install the plugin mypads, so you'll have just etherpad-lite.

So, I think we could get rid of etherpad_lite, as this package isn't maintained anymore.

alexAubin:
Soooo, is that okay for everybody ? (@JimboJoe ?)

PR #631, opened 2018-12-03
add zabbix

Mickael-Martin:


community.json
```diff
@@ -1730,6 +1730,13 @@
         "revision": "795449cb43ccdba949c814746d6ea90eee0a7259",
         "state": "working",
         "url": "https://github.com/aymhce/yunohost_ynh"
+    },
+     "zabbix": {
+        "branch": "master",
+        "level": 2,
```
alexAubin: Please don't specify the level manually, it's meant to be the result of automatic test by the CI

community.json
```diff
@@ -1730,6 +1730,13 @@
...
+        "revision": "9386c5b2",
```
alexAubin: I think you need to put the full commit id here, not just a short one ?

Mickael-Martin: Gitlab works with small ids like this : 
https://framagit.org/Mickael-Martin/zabbix_ynh/commits/master

community.json
Mickael-Martin: Sorry, who make this test ? me ? you ? using https://github.com/YunoHost/package_check ?

frju365: Nobody. It's the CI who did automatic tests : https://ci-apps.yunohost.org/ci/

frju365: You have just to remove this line.

alexAubin:
Thanks !

Mickael-Martin:
Thank you too !

Mickael-Martin:
sorry, why the app no appears on the community list ?

Mickael-Martin:
in this list : https://app.yunohost.org/community.json

alexAubin:
Thanks for reporting the issue.

I manually launched the build of the list (the community.json in this repo and the community.json you are pointing are two different files for various reasons)

I found this error : 
```
-> Error: unable to decompress object from https://framagit.org/Mickael-Martin/zabbix_ynh.git/objects/93/86c5b2 : Error -3 while decompressing data: incorrect header check
```

And it looks like this URL is indeed not working. Do you have an idea of the url we should use ?

Mickael-Martin:
what is your command/code to obtain that ?
If you want try, I open a pushed mirror repo on https://github.com/Mickael-Martin/zabbix_ynh

I see the same issue with : 
  "wifiwithme": {
        "branch": "master",
        "level": 0,
        "revision": "14e01a3981bde7192e100b1f9cfa50e95d6cb89b",
        "state": "notworking",
        "url": "https://code.ffdn.org/ljf/wifiwithme_ynh"
},

alexAubin:
To explain a bit what's going on, basically the "built" json file shall include the manifest of the app, among other things.

The relevant piece of code is here : https://github.com/YunoHost/apps/blob/master/list_builder.py#L252-L255

Sounds like there is indeed an issue with another app hosted on framagit (covoiturage_ynh) though I'm kinda sure it worked at some point ... Maybe the Gitlab API changed. Anyway, we shall investigate how to contact the API 'cause that's an important piece if we hope to move away from Github someday. The tricky part is that some other apps (e.g. wifiwithme) might be hosted on a Gogs with yet another API ...

alexAubin:
I'm in the process of fixing the issue, but in the meantime, please consider changing the 'id' inside your `manifest.json` to just `zabbix` instead of `ynhzabbix` ... Otherwise people will see your app being called `ynhzabbix` which makes less sense than just `zabbix` :s

Mickael-Martin:
Done ! Thanks !

alexAubin:
Thanks !

Though please note that this won't be effective until the `revision` key is updated in `community.json` :sweat_smile: . (Yes, that's a lot of details to handle ..)

Nowadays we usually set the `revision` to `HEAD` to avoid updating it each time a new stable version is release. Though that means that you should probably have a `testing` branch for commits that are not suitable yet for production. See the note about HEAD here : https://github.com/YunoHost/Apps/#how-to-add-your-app-to-the-community-list

frju365:
revision commit is : 9386c5b28e44ced928e886c00685d0f9a8486620 not the small one (we need the entire commit revision number - SHA number which is generated in fact)

alexAubin:
(The short one works as well as the long one)

frju365:
ok didn't know :P I prefer the complete one, as others PR has it.

PR #634, opened 2018-12-06
Adds Pilea

SimonMellerin:
Hi,

I'm new here, I developed an little app to monitor electricity consumption via Linky. 

Maybe it can interset some people :)


maniackcrudelis:
That looks very interesting !
I'm going to try it this afternoon

PR #637, opened 2018-12-12
Made SHSD level 3 and working (according to CI)

flesueur:


community.json
```diff
@@ -1395,9 +1395,9 @@
     },
     "shsd": {
         "branch": "master",
-        "level": 0,
+        "level": 3,
```
alexAubin: Thank you for your contribution, but please do not change level manually ... this is meant to be set automatically from CI results. Otherwise this number can't be trusted :/.

alexAubin: Thanks !

jgpa3qa

PR #642, opened 2018-12-17
Fix some inconsistencies in app names

alexAubin:
Those are some fixes for the name of some apps, because it's quite annoying to have so many inconsistencies between the manifest id, the key in community.json, and the name of the git repo ...

Those should hopefully make a bit more sense ... It should supposedly be a benign change since the key in community.json is anyway a non-reliable info ... (Ideally in the longterm I wish we could just make it disappear and just use a list of dict instead of a dict of dict...)

(Note that some renaming here are to sync the name with the manifest id, which is some cases is different from the name of repo itself (sigh))

### DIFF
--- a/community.json
+++ b/community.json
@@ -81,7 +81,7 @@
         "maintained": false,
         "revision": "49fb42612bb37f44960b93b872f3523fa615f203",
         "state": "inprogress",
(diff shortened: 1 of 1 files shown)


alexAubin:
Added another fix for case synchronisation (e.g. Youtube-dl-WebUI <-> youtube-dl-webui) which should also hopefully be a benign change...

PR #646, opened 2018-12-21
Update commit from some apps, and set them maintained

frju365:
I will update and maintain some apps which where forgotten ;)


frju365:
Well, for most of them I already maintain it :P It's just mediawiki I reset it. Roadiz is set as unmaintained for upstream reason

PR #656, opened 2019-01-13
Adding WriteFreely Application

yalh76:


maniackcrudelis:
If we remove this app from the list, for sure it's never going to be fixed...
This app is not maintained nor working, no one is supposed to install it.

Anyway, if you want to remove this app, feel free to open a pull request. But do not add your app while you remove another one from someone else...

yalh76:
Ok thanks for your help and response.
For the removing of an app, adding another: I didn't intend to do that... I'm a newbie on git.
I will do another pull request ^^

PR #659, opened 2019-01-15
change url repo for zabbix and set revision to head

Mickael-Martin:
Hi !

Just a change of URL after migration of my app to community repo.
I set revision to HEAD to avoid pull-request at any update.


Mickael-Martin:
Thank you very much @JimboJoe !

PR #662, opened 2019-01-20
update of writefreely level

yalh76:
According:
- https://ci-apps.yunohost.org/ci/job/1651
- https://github.com/YunoHost-Apps/writefreely_ynh

### DIFF
--- a/community.json
+++ b/community.json
@@ -1711,6 +1711,7 @@
     },
     "writefreely": {
         "branch": "master",
+        "level": 7,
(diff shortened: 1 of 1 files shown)


frju365:
Normally, we don't touch level number. The IC will normally open its Pull Request. I will add it in the PR #661 (for other time, it's the same : when a PR is opened, we can change it, otherwise, you will have to wait the CI PR :-) )

PR #673, opened 2019-01-29
Adding distbin

yalh76:
image

I don't know how to move the repository to YunoHost-Apps


frju365:
Anyway, normally, you don't have to set the level. It's the CI and only the CI which can set this level.

PR #677, opened 2019-02-01
Add support for featured and high quality apps

maniackcrudelis:
Replace https://github.com/YunoHost/apps/pull/655 with a tag in the community list instead of a new list.
Related to https://github.com/YunoHost/yunohost-admin/pull/225 to manage the tag in the app list.

For all who'd worried that I really modify python code. Don't worry, I wasn't alone to do it.

The discussion about this feature is [here](https://forum.yunohost.org/t/about-community-and-official-apps/6372). Nothing has been decided yet, still I go forward to propose some concrete work.
I especially would like to have the point of view of the @YunoHost/apps group, since we would be concerned as we would have to give a quick review to each PR on those apps.

---

EDIT

During our last meeting it was decided:

- In a mid-long term, get rid of Official and Community list, and merge into one unique list.
Let's say simply "One list to rule them all, One list to find them; One list to bring them all and [...]." Don't remember the end of this sentence... Never mind...
- Add a status of the maintained state of an app. Like the way Debian do it. ‘Orphaned’, ‘Request for Adoption’, ‘Request For Help’.
- Keep 'Featured apps' as apps highlighted for editorial reasons. Maybe via a poll on the forum, not defined yet.
- Adapt level 8 for high quality apps. Validated by Apps group.
- Modify level 4, to not block apps anymore under level 3 because of ldap or SSO.

Related PR:
Those PR should be merged in that order to not break anything.
- This very PR should be merged first.
- ~Add High Quality tag for some apps [#678](https://github.com/YunoHost/apps/pull/678)~
- Modify levels 4, 6 and 8 [#52](https://github.com/YunoHost/package_check/pull/52)
- Add YEP 1.12 - Follow example_ynh [#935](https://github.com/YunoHost/doc/pull/935)
- Update packaging_apps_levels.md [#927](https://github.com/YunoHost/doc/pull/927)
- Add support for featured community apps [#225](https://github.com/YunoHost/yunohost-admin/pull/225)
- Merge Official and Community into Apps [#694](https://github.com/YunoHost/apps/pull/694)
- Migrate to apps.json [#5](https://github.com/YunoHost/yunorunner/pull/5)
- Use apps.json as default list [#665](https://github.com/YunoHost/yunohost/pull/665)
-  Migrate to apps.json [#666](https://github.com/YunoHost/yunohost/pull/666)
- Get apps from apps.json instead of community and official [#13](https://github.com/YunoHost/CI_package_check/pull/13)

TODO list:
[...]

PR #678, opened 2019-02-01
Add High Quality tag for some apps

maniackcrudelis:
Following https://github.com/YunoHost/apps/pull/677, here a proposition for some apps which can be the first to be tag as Featured. So we could start the new feature.
That's only a proposition to begin with this new tag.

Those apps are only apps maintained by members of apps group, considering that one of the requirement is to accept that the group will have a look to each PR. It seems to me that it would be better to start with apps from the group itself.
Obviously one would notice that most of them are my own apps. Indeed, I don't need a long and harassing review to know what the status of my own apps. And I know that they already respect the requirements.
Also, I intentionally removed apps which used too much factorization in _common.sh, since I'm again such way to code apps here. I'm still in favor of keeping scripts clear and simple, and to stick to the template.

I notice also that [cachet](https://github.com/YunoHost-Apps/cachet_ynh), [garradin](https://github.com/YunoHost-Apps/garradin_ynh) and [hubzilla](https://github.com/YunoHost-Apps/hubzilla_ynh) miss a testing branch.

This PR is only a proposition to start a list of Featured apps. It could be discussed.
And, finally, this PR has only a meaning if https://github.com/YunoHost/apps/pull/677 and the global idea of this tag is accepted.

**This PR should be merged after https://github.com/YunoHost/apps/pull/677**

### DIFF
--- a/community.json
+++ b/community.json
@@ -53,6 +53,7 @@
     "archivist": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/archivist_ynh"
@@ -113,6 +114,7 @@
     "cachet": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/cachet_ynh"
@@ -347,6 +349,7 @@
     "fallback": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/fallback_ynh"
@@ -444,6 +447,7 @@
     "friendica": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/friendica_ynh"
@@ -472,6 +476,7 @@
     "garradin": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/garradin_ynh"
@@ -644,6 +649,7 @@
     "hubzilla": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/hubzilla_ynh"
@@ -775,6 +781,7 @@
     "leed": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/leed_ynh"
@@ -838,6 +845,7 @@
     "lutim": {
         "branch": "master",
         "level": 7,
+        "featured": true,
         "revision": "HEAD",
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/lutim_ynh"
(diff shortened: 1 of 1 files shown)


Josue-T:
> Also, I intentionally removed apps which used too much factorization in _common.sh, since I'm again such way to code apps here. I'm still in favor of keeping scripts clear and simple, and to stick to the template.

Why ?? I don't really think that duplicating the code is the best idea...

maniackcrudelis:
I was expecting that, especially from you :)
The reason is quite simple, our apps scripts have a standard template, which make packages so easy to create.
That template is also what makes reviews and fixes easy and painless, we don't have to navigate between multiple files to find what going on.
Obviously, as developers we all have learned that it's better to factorize the code as much as possible to reduce maintenance. But here, we're in a specific situation, where we often have to review code from others devs. Really often ! And there's nothing worse than to have to understand how the code is built to be able to review, especially as we're supposed to have a clear template, used by everyone.
As I have often reviewed packages, as well as I've fixed them. I know perfectly how painful it is to have to jump from file to file and from function to function to retrace the (initially) straight flow of the script. Even more when I have to use git blame, it's then a real pain in the ass since git blame doesn't work with multiple files.
I loose an incredible amount of time struggling with that kind of code.

So for all those reasons, I keep thinking that factorizing the code in app packages is not a good idea. And honestly, we're not changing those parts of the code every day, duplicating them is just better for all other devs who will have to review your code.

Josue-T:
I still think that there are no reason to discriminate the factorized apps. I see nowhere a specification for app packaging which say that code factorization is not authorized. All software factorize the code, so I don't see any reason to disallow this practice for this code. 

@YunoHost/apps what do you think about that ?

JimboJoe:
I'm not opposed myself to code factorization. Considering example_ynh provides an example/guide/template without using factorization, I do think that people using factorization will have a greater chance to know what they're doing, and I'm not in favour of opposing to that practice.
We'll still have to review the changes, but the main idea of the process change is based on the fact we don't have much time, so we'll have to be a little more trustful toward featured apps maintainters I guess :wink:

maniackcrudelis:
I had to work on our official apps, and I'll continue to do so for next days/weeks.
That's a fucking hell to have to work on apps with factorization. You can't rely to the template, you have to go through multiple files and functions to find what's going on.

I wouldn't oppose if packagers were working on their own only on their apps. But that's not what happens. We have really often to work on the code of someone else. Using a template is really easier when you have to check an app and find what's not working or what to update.
I was able to work really fast on apps that respect the template, but I loose a lot of time and serenity figuring out how some apps were working.
That's for the first part, about packagers who know how to deal with bash. But don't want to loose time figuring out how the fuck another guy made his package. (Where we're supposed to have a clear and easy template)

The other part we have to consider is all our newbies packagers who start or will start.
With a clear and easy template, any new packager willing to help will be able to take back any unmaintained package and fix it or to propose any update on a maintained package.

If you really prefer to code every one on its side, I can also stop restraining myself when I code an app and just code as I want. Like some of my helpers.
But I don't think it would be really clever to just think about how it could convenient for ourselves to code as we want.
We have to think globally about our packages.

Otherwise I could also stop doing the shitty job of rewriting all those shitty apps with non standard code, and let you loose your time trying to have easy apps to maintain.
Or just stop trying to have a clear template for everyone, stop trying to have apps that are easy to build and maintain.
If you don't care, why should I !?

maniackcrudelis:
Sorry if I'm a "little bit" upset about that. But I really do loose a lot of time and serenity dealing with those kind of apps. And that happens too often.

maniackcrudelis:
Considering the new template for validation of such apps, https://github.com/YunoHost/apps/pull/677/commits/abdcf2b7b91a22c73c3feab7d71178fe4d22eefb (not yet merged)

I think it would be better to advertise this new tag as soon as the main PR will get merged. And then let packagers ask for the tag.

So, I close this PR. Feel free to reopen if you disagree.

PS: That doesn't change anything about what had been said above...

PR #684, opened 2019-02-05
filebin removal

yalh76:
I propose to remove filebin:

- Yunohost Package no more maintained
- Upstream application is not maintained since more than 5 years : https://github.com/sebsauvage/ZeroBin
- The maintainer of the upstreamapplication advises to use PrivateBin https://sebsauvage.net/wiki/doku.php?id=php:zerobin
- PrivateBin is already a Yunohost package named: zerobin => https://github.com/YunoHost-Apps/zerobin_ynh

### DIFF
--- a/community.json
+++ b/community.json
@@ -360,14 +360,6 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/ffsync_ynh"
     },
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Agree

PR #693, opened 2019-02-21
Add mindmaps

justAyTE:
I tested it with package_check only the restore script failed but it seems like an error from the core

https://paste.yunohost.org/ripofasose.sql

### DIFF
--- a/community.json
+++ b/community.json
@@ -933,6 +933,12 @@
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/minchat_ynh"
     },
+    "mindmaps": {
+	"branch": "master",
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Why do you get the source from your own repo ? https://github.com/YunoHost-Apps/mindmaps_ynh/blob/master/scripts/install#L88

zamentur:
It's because the apps contains some trackers a pull request on upstream has been made.
https://github.com/drichard/mindmaps/pull/84

Yesterday he didn't have time to learn how to make a patch and import the upstream with the standard way.

PR #701, opened 2019-02-23
Readme Apps - Images missing - Fix issues #1312

toitoinebzh:
This PR fix https://github.com/YunoHost/issues/issues/1312


alexAubin:
Personally I would just choose to remove the images ... I don't think these are so useful and it's a pain to maintain in the long-term / nobody thinks about it

toitoinebzh:
Better like this :)

JimboJoe:
Agreed, images in that README are really not necessary in my opinion.

maniackcrudelis:
Totally agree with that !

toitoinebzh:
should I do something to merge this PR ? (noob inside)

alexAubin:
Yolomergin' because there are 4 approvals :wink:

toitoinebzh:
thank you :smile:

PR #721, opened 2019-03-13
Update SPIP

kay0u:


maniackcrudelis:
Does that mean you're going to maintain both spip2 and 3 ?
Anyway for now, I don't think that would be a good idea to put those 2 apps at HEAD, considering the last commits are 2 years old.

kay0u:
In fact that don't change anything for now, the commit in this list corresponding to HEAD on each repo.

I'll add a script migration for spip2 => spip. But as there are missing some informations inside settings in spip2, I'm wondering... Maybe I should in the upgrade script of spip 2 add these informations.

People who want to migrate to spip would have to upgrade spip2 before...

maniackcrudelis:
If the upgrade of spip2 works, there's no problem to ask for an upgrade of it before the migration.

kay0u:
Ok

PR #694, opened 2019-02-21
Merge Official and Community into Apps

maniackcrudelis:
Following https://github.com/YunoHost/apps/pull/677

This PR propose to merge Official and Community into one list simply named "Apps"
Official apps are tagged as `working` instead of `validated` and get by the way the flag `high_quality`. Because there were Official apps so far.

**This PR should be merged after https://github.com/YunoHost/yunohost-admin/pull/225**

PR #727, opened 2019-03-21
list_builder.py backward compatibility : split apps-build.json to community and official

alexAubin:
So here's finally the small PR I promised 2 weeks ago ;P

Running list_builder.py on apps.json will (in addition to keeping `apps-build.json`) produce `official-built.json` and `community-built.json`

Since I needed it, I also made a very small one-time script to merge official and community to produce apps.json (which supersededs a part of #694 )


maniackcrudelis:
Nice

That's indeed better to merge this PR instead of #694

maniackcrudelis:
Can be merged whenever it's ready.

alexAubin:
Mergin as agreed with Maniack

PR #728, opened 2019-03-24
Update app's level

maniackcrudelis:
Wait for new tests for
- [discourse](https://ci-apps.yunohost.org/ci/job/2402)
- [minidlna](https://ci-apps.yunohost.org/ci/job/2334)

### DIFF
--- a/community.json
+++ b/community.json
@@ -84,6 +84,7 @@
     },
     "blogotext": {
         "branch": "master",
+        "level": 7,
         "maintained": true,
         "revision": "HEAD",
         "state": "working",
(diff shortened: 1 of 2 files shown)


alexAubin:
Lots of regression eh :s Are those kinda understood ? (I'm guessing it might be related to the fixes in package linter ?)

I also see some improvements which might be related to the change of definition for level 4 ?

maniackcrudelis:
Mainly, regression are because of the linter, for apps stuck at level 4.
For those which have decrease from 5 to 3, that's because of the modification of the level 4 for backup/restore.
Some other simply doesn't work anymore.

Except for discourse and minidlna, nothing unexpected.

PR #732, opened 2019-04-05
Adding anfora, mobilizon and reel2bits

yalh76:


alexAubin:
photo

yalh76:
Nop, not yet ^^
But three projects I want to work on. Upstreams are all in early alpha.
But they won't be difficult to package, one use elixir (copy/paste from pleroma) and the two others python3

It's new activitypub services not already implemented on Yunohost ^^

PR #733, opened 2019-04-05
Update community.json

liberodark:
Add Adguard


kay0u:
Hi, thank you for this app.

But please, don't provide the "level"? This field is added by our runner after tests.

PR #739, opened 2019-04-06
Update app's level

maniackcrudelis:


kay0u:
We still have a problem with pgadmin: https://ci-apps.yunohost.org/ci/job/2501

maniackcrudelis:
For this app there's indeed a strange problem!
I tested the app on my own yesterday evening, and there's indeed a issue. The app failed all tests and got a level 0 on my tests.
I don't know why the CI shows success for all tests. Anyway I think this app should be retested and fixed.

Josue-T:
I found the issue on the app side.

It should be nearly fixed

PR #740, opened 2019-04-13
Notworking apps

maniackcrudelis:


apps.json
```diff
@@ -1703,7 +1703,7 @@
         "branch": "master",
         "level": 0,
         "revision": "HEAD",
-        "state": "working",
+        "state": "notworking",
```
Josue-T: Well, I think there are a little problem.

The name of this app is "sogo" but the url is "stream_ynh"...

I think this url should be https://github.com/YunoHost-Apps/sogo_ynh
And normaly this app is working

PR #748, opened 2019-04-21
switch z-push to Head

yalh76:


maniackcrudelis:
Before switching this app to HEAD, it have to be maintained.
Meanwhile, you can switch to the last commit if you want.

yalh76:
> 
> 
> Before switching this app to HEAD, it have to be maintained.
> Meanwhile, you can switch to the last commit if you want.
I've switched to the last commit
Well, in that case, can you put me as maintainer of z-push_ynh and give me rights on the repository ?

PR #759, opened 2019-05-04
Add CodiMD

lapineige:
Add CodiMD to the list: https://github.com/YunoHost-Apps/codimd_ynh

Note: I'm not the packager of this app. @liberodark @frju365 please add your reaction if needed.

I took the liberty to add the "HEAD" option (which means auto update but that development versions should be in a testing branch).
I don't really know the package status so I have chosen `in progress`.

### DIFF
--- a/community.json
+++ b/community.json
@@ -179,6 +179,12 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/chtickynotes_ynh"
     },
+    "codimd": {
+        "branch": "master"
+        "revision": "HEAD",
+        "state": "inprogress",
+        "url": "https://github.com/YunoHost-Apps/codimd_ynh"
+    },
     "coin": {
         "branch": "master",
         "level": 0,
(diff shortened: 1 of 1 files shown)


community.json
```diff
@@ -179,6 +179,12 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/chtickynotes_ynh"
     },
+    "codimd": {
+        "branch": "master",
+        "revision": "HEAD",
```
kay0u: I'm more in favor of a fixed commit instead of HEAD, there is no testing branch (you can use sha1 of the current HEAD revision ;-))

lapineige: That's why I asked @liberodark and @frju365 to confirm if they are ok with HEAD, which means creating a testing branch.
It's easier for packagers with HEAD.

frju365: Normally : fixed : https://github.com/YunoHost-Apps/codimd_ynh/issues/6

PR #760, opened 2019-05-04
Update Chtickynotes - Set maintained to false + update version

lapineige:
This app is no longer maintained : https://github.com/YunoHost-Apps/chtickynotes_ynh/issues/7

### DIFF
--- a/community.json
+++ b/community.json
@@ -175,8 +175,9 @@
     "chtickynotes": {
         "branch": "master",
         "level": 0,
-        "revision": "94575a0db73e0ba1028882c625ebb363b0450f01",
+        "revision": "HEAD",
         "state": "notworking",
+        "maintained": "orphaned",
         "url": "https://github.com/YunoHost-Apps/chtickynotes_ynh"
     },
     "coin": {
--- a/community.json
+++ b/community.json
@@ -175,7 +175,7 @@
     "chtickynotes": {
         "branch": "master",
         "level": 0,
-        "revision": "HEAD",
+        "revision": "94575a0db73e0ba1028882c625ebb363b0450f01",
         "state": "notworking",
         "maintained": "orphaned",
         "url": "https://github.com/YunoHost-Apps/chtickynotes_ynh"


maniackcrudelis:
Why put the commit to HEAD ? Especially if no one maintain it.

lapineige:
At least it doesn't require to do a PR to this community.json file for every (little) change ?

Even if I'll not maintain it, I might try to improve it a bit, maybe with an upgrade of the source.

maniackcrudelis:
Using HEAD implies that the maintainer is working on a testing branch.
Otherwise, it would be better to update the commit when something concrete (and working) is done.

lapineige:
> Using HEAD implies that the maintainer is working on a testing branch.

Oh I forgot to mention @chtixof… I wanted to ask @chtixof if it was ok to do that, and I'm ok to setup this branch if I work on the package.

chtixof:
Thanks @lapineige for mentionning me. Do it for the best.

lapineige:
Then @maniackcrudelis I'm fine to be working with testing + master branches.

maniackcrudelis:
Do that means @lapineige that you're willing to take back the maintenance of this app ?

lapineige:
No, but I may take some time to update it once.

maniackcrudelis:
So, in that case, there's no reason to put the revision to HEAD.
An upgrade would be surely welcome, but then you just have to update the revision to the last commit when you're sure it works.

frju365:
@lapineige : That @chtixof works on app doesn't prevent you from being maintainer too. :-) I mean that two maintainers is better than one.

lapineige:
I don't have the time to maintain it.

yalh76:
This PR seems out of date, maybe we can close it ?

lapineige:
It's not out of date as the package status didn't change.

maniackcrudelis:
As well the PR hasn't been fixed as asked.
So whether this PR is going forward, or there's no need to keep it open.

lapineige:
Well, not using HEAD just sounds like creating unnecessary work in my opinion (open a new PR each time the package change), but it's done.

maniackcrudelis:
Using HEAD implies the package is maintained and the maintainer is working on a testing branch.
So, the master branch is updated and made available for everyone ONLY when it's tested and it works.

Using HEAD here will made the package updated for all users at each commit, whether the commit is tested or not.

lapineige:
And that what I'm was planning to do, not merging PR without testing it.
The package is broken anyway right now…

PR #761, opened 2019-05-06
add armadietto_ynh to Yunohost apps list

bencharp:


apps.json
```diff
@@ -110,6 +110,12 @@
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/archivist_ynh"
     },
+    "armadietto": {
+        "branch": "master",
+        "revision": "HEAD",
```
maniackcrudelis: Please consider though that using HEAD implies you're working on a testing branch, and release only stable work on master.

bencharp: Testing branch created!
Branche testing créée !

maniackcrudelis: Thanks

PR #763, opened 2019-05-09
Set 'redirect' as working + use HEAD commit

alexAubin:
The current commit dates back from 2017 :s Also recent test shows that this app is at least level 1, so setting as working

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1530,8 +1530,8 @@
         "branch": "master",
         "level": 0,
         "maintained": false,
-        "revision": "2230d41e8b788fa35fe3bf0769b71ab24a002e95",
-        "state": "inprogress",
+        "revision": "HEAD",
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
This app doesn't seem to be maintained, not sure that's a good idea to set the revision to HEAD.

alexAubin:
Zbleh idk I guess I can declare myself as the maintainer then, it's a small app ¯\\\_(ツ)\_/¯

I'll make a PR

alexAubin:
Sooo c.f. https://github.com/YunoHost-Apps/redirect_ynh/commit/238eeb8f57b20a23a5de7173f612bf3364d39d09

More work to come on this app as I discovered that 'is_public=no' doesn't work for this app due to nginx subtilities

PR #790, opened 2019-06-04
removing apticron

yalh76:
already included in YunoHost

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -98,13 +98,6 @@
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/anfora_ynh"
     },
-    "apticron": {
-        "branch": "master",
-        "maintained": false,
-        "revision": "4fb8696bd1499caf9d7177a7bbb0ea14592331a9",
-        "state": "inprogress",
-        "url": "https://github.com/ldidry/apticron_ynh"
-    },
     "archivist": {
         "branch": "master",
         "level": 7,


maniackcrudelis:
Actually not, just tried on a VM, apticron isn't obviously installed.
And not either part of YunoHost dependencies, https://github.com/YunoHost/yunohost/blob/stretch-unstable/debian/control

However, I don't think we could say that's a real YunoHost app package, https://github.com/ldidry/apticron_ynh/blob/master/scripts/install
Just an apt install. And I think for users who know what apticron is, it's easier to apt it yourself.

So, agree to remove it.

yalh76:
> 
> 
> Actually not, just tried on a VM, apticron isn't obviously installed.
my bad, apticron is installed in the vms I use because I have installed [unattended-upgrades_ynh](https://github.com/YunoHost-Apps/unattended_upgrades_ynh) in the master image ^^

kay0u:
Remove a package just because it's not packaged like standards is not a good argument for me.

I'm more in favour to repackage it (I can do it if needed)

yalh76:
> 
> 
> Remove a package just because it's not packaged like standards is not a good argument for me.
> 
> I'm more in favour to repackage it (I can do it if needed)

oups... I did that: https://pad.lqdn.fr/p/yunohost-18-06-2019

kay0u:
Remove a non working/in progress app unmaintained is for me not a good idea too (but we can obviously discuss that the next meeting!).

The point is, the list provide a kind of visibility for unmaintained app, and can motivate someone to take care of it.

We already have level to warn that the app is not well packaged, and like 2 or 3 to tell that this app could break something

yalh76:
> 
> 
> Remove a non working/in progress app unmaintained is for me not a good idea too (but we can obviously discuss that the next meeting!).

I was thinking like you, but aleks asked me if it's possible to add all the not working/in progress app to the https://github.com/yalh76/apps_poll. I done it going to each repository.... and there is a lot of crap... sorry to say that... some 4 years old apps.

Some example:
- [jitsi](https://github.com/YunoHost-Apps/jitsi_ynh) I will work on it but the more I try to apply example_ynh the more I think I will replace all the repository by example_ynh and redo all from scratch
- [subsonic](https://github.com/drfred1981/subsonic_ynh) level 0/4 years old/inprogress / already replaced by a level 7 fork airsonic_ynh
- [mycryptochat](https://github.com/YunoHost-Apps/mycryptochat_ynh) level0 / notworking / 4 years old , [upstream app](https://github.com/HowTommy/mycryptochat/) : 4 years, [upstream app maintainer blog](http://dev.howtommy.net/index.php?d=2015/06/01/08/26/35-#id530505) 4 years without a post.....

But it's more my point of view, and really not an important point to discuss as it's just a few lines in apps.json

maniackcrudelis:
> Remove a package just because it's not packaged like standards is not a good argument for me.

Not even that it's not packaged by following standard, it's not even a real app package.
Just an apt install in an install script and an apt remove in a remove script.

kay0u:
After rethinking of that, I changed my mind. I agree to remove it.

maniackcrudelis:
Let's merge in 3 days

PR #795, opened 2019-06-22
Remove unmaintained apps which are a ~redundant with my_webapp

alexAubin:
While inspecting the list of apps yesterday, found out about these two apps which really sounds like a duplicate work of the custom webapp things (my_webapp and multi_webapp) ... So proposing to remove them from the list because that's just confusing and redundant

(Also I wish I would understand if there's really a need behind the my_webapp and multi_webapp - can't we add FTP support in multi_webapp as well ?)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1743,14 +1743,6 @@
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/ssh_chroot_dir_ynh"
     },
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Old and unmaintained, let's remove them.

Let's merge in 3 days.

PR #796, opened 2019-06-22
CodiMD, NodeRed and ShadowSocks were added to community.json instead of apps.json

alexAubin:
We should really consider removing community.json and official.json to avoid that kind of confusion :/


kay0u:
We must not add anything to these lists, but we cannot delete them, and we plan to keep them at least 6 months after the integration of apps.json into YNH officially (with a migration tool)

PR #798, opened 2019-06-23
Switching inprogress apps not changed since 6 months to notworking

yalh76:


maniackcrudelis:
So many apps...
Looks like a good idea though.

Let's merge in 3 days.

PR #799, opened 2019-06-29
Update app's level

kay0u:
- Cachet
- Libreerp


maniackcrudelis:
Thanks @kay0u for being here taking care of that !

PR #803, opened 2019-07-09
Rename openvpn to vpnserver

alexAubin:
Because it's been like 6 months I'm mumbling about this so I finally took the dictatorial decision to rename it because I'm so mad about the ambiguity of the name (vpnclient is also based on openvpn so)


Psycojoker:
Ok with this decision, everything that reduce support load is good to take.

alexAubin:
Yolomergin' ... (cc @zamentur )

PR #809, opened 2019-08-01
All these apps are not "working", shouldn't have a level > 0

alexAubin:


kay0u:
:thinking: Uuh, a non working app shouldn't have any "level" if I remember well.

alexAubin:
Yea, we can also remove those entirely ¯\\\_(ツ)\_/¯

maniackcrudelis:
Those levels were set before those app were put as not working (or in progress for some by the way)
Is it really a problem to have those ?

alexAubin:
Not really since most codes reading this file will / should assume level to be 0 (or None) if the app ain't working. Still, I wouldn't be surprised that some users read this raw file and think that "oh this app is notworking but still is level 3 so maybe I can install it"

PR #820, opened 2019-08-28
Change into inprogress

zamentur:
We are currently discussing if we have the right to package CODE. So i suggest to change into inprogress until we have the answer.


yalh76:
Just for information, there is also: https://github.com/YunoHost-Apps/collabora_ynh with a PR ready for a level 7

alexAubin:
Shall we merge this then ?

PR #821, opened 2019-09-16
Update app's level

kay0u:
We really need to avoid resetting the level of an non working app (or resetting it to 0... I'm tired of doing it every week)

@maniackcrudelis @alexAubin any idea how to do that?

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -31,7 +31,7 @@
     },
     "adhocserver": {
         "branch": "master",
-        "level": 2,
+        "level": 0,
         "maintained": false,
         "revision": "d1a728b9b99608bac69b55372cddf1aa3f4a5557",
         "state": "notworking",
@@ -361,7 +361,7 @@
     },
     "docker-registry": {
         "branch": "master",
-        "level": 3,
+        "level": 0,
         "maintained": true,
         "revision": "HEAD",
         "state": "notworking",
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
The question would be do we really need to set the level to 0 in the first place...
A not working app is not working... Whatever the level.

If we really want that.
Then, there's 2 solutions. Change the code of the CI to set level to 0 itself if the app is not working. Or modify the list of levels and set the levels to 0 for those apps

kay0u:
Somewhere in this file: https://github.com/YunoHost/CI_package_check/blob/master/auto_build/update_level_apps.sh

kay0u:
The thing is more that the runner shouldn't set any level for an app not tested anyway

maniackcrudelis:
> The thing is that the runner shouldn't set any level for an app not tested anyway

It doesn't, all apps with a level have been tested. But those apps changed their status since.

kay0u:
That's true, but since these apps have changed their status, we shouldn't rewrite their levels.

kay0u:
What are we doing ?

kay0u:
Can I delete the old logs of `/home/CI_package_check/logs/` in the CI if necessary?

kay0u:
I close this PR because radicale and synapse are coming back to their level:-)

PR #824, opened 2019-09-24
Update flarum_ynh

tituspijean:


tituspijean:
By the way, with this commit I removed the "level" line, and it has not come back yet. Shouldn't it be automatic?

kay0u:
We're working on it now, you'll probably get it next week. :-)

PR #827, opened 2019-10-01
Adding OpenProject in the app list

moutonjr:
OpenProject is still ongoing, the bash code should be neat.

See https://forum.yunohost.org/t/app-integration-of-openproject/9141/2 for details
and https://www.openproject.org/ for the software.


kay0u:
Thank you very much for this contribution!

PR #833, opened 2019-10-06
force level if not working

kay0u:
If an app have the state "notworking", then force the level to 0.

Should I do the same for "inprogress"?

### DIFF
--- a/change_level.py
+++ b/change_level.py
@@ -32,7 +32,7 @@ if __name__ == '__main__':
         
     if app_list[app_id]["state"] == "working":
         app_list[app_id]["level"] = int(level)
-    else:
+    elif "level" in app_list[app_id]:
         print "Warning: app '%s' is currently not marked as working, the level is removed" % app_id
         del app_list[app_id]["level"]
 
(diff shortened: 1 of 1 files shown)


kay0u:
Bump

alexAubin:
Uuuuh can you confirm that it's what's actually used to update app levels ?

Code-wise, to me it should be more something like : 

`if app["state"] != "working":
    del app["level"]`

so that it entirely deletes the level key compared to setting it to 0 (which suggests a test was made) but that's kind of a detail, dunno if we really care so much about it, the important bit being that it's not a value higher than 0...

kay0u:
https://github.com/YunoHost/CI_package_check/blob/594a25480a7af39bf8cf9b853955df267d73a857/auto_build/update_level_apps.sh#L36

From what I read, yes it is.

You probably right, I'll change that, moreover the status `inprogress` should remove level too.

kay0u:
Should be ok now

change_level.py
```diff
@@ -29,7 +29,11 @@
     if not 0 <= int(level) <= 10:
         print "Error: app level must be between 0 and 10, it's '%s'" % level
         sys.exit(1)
-
-    app_list[app_id]["level"] = int(level)
+        
+    if app_list[app_id]["state"] == "working":
+        app_list[app_id]["level"] = int(level)
+    else:
```
alexAubin: Hmm I think that might trigger an issue if the level key does not exist :P

```suggestion
    elif "level" in app_list[app_id]:
```

PR #841, opened 2019-10-19
Modify level

kay0u:


kay0u:
Yep, nothing has changed on these repo, and there is no reason of a drop. In addition, after the CI was restarted, levels goes back to 7

PR #842, opened 2019-10-25
Update app's level

kay0u:
~Waiting for framaforms and grav~


maniackcrudelis:
Is it ok to remove all those level 0 ?

kay0u:
As not working apps (or in progress) are not tested, is it legit to have a level on these apps?

PR related: https://github.com/YunoHost/apps/pull/833

We can discuss that if you are not agree, and revert/change this PR to force the level 0

maniackcrudelis:
I didn't remember what was the decision about that.
It was just to know before removing all those levels

Still, to have a level 0 for a not working app seems relevant. As, the app does not work.

kay0u:
Then we should write level 0 for all applications that are not working / running. For the moment, the CI is content to [rewrite the level of the application](https://github.com/YunoHost/CI_package_check/blob/c714ae81da4815b3b922fae9bd9b7fbf78db5465/auto_build/update_level_apps.sh#L39) that has already been tested (because it was working?) but not anymore. We don't know if the application is level 0, we don't test them.

kay0u:
So, merging or keeping level 0 for apps not working?

maniackcrudelis:
Just do as you usually do since I'm not here.
I was just noticing that strange commit with so much apps with their level removed.

kay0u:
Ok, hope to see you back soon!

PR #845, opened 2019-11-04
Update app's level

kay0u:
~Waiting for framaforms, libreerp and mantis~

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -518,7 +518,7 @@
     },
     "framaforms": {
         "branch": "master",
-        "level": 7,
+        "level": 2,
         "maintained": true,
         "revision": "HEAD",
         "state": "working",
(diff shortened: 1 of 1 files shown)


apps.json
```diff
@@ -518,7 +518,7 @@
     },
     "framaforms": {
         "branch": "master",
-        "level": 7,
+        "level": 2,
```
kay0u:
```suggestion
        "level": 7,
```

apps.json
```diff
@@ -916,7 +916,7 @@
     },
     "libreerp": {
         "branch": "master",
-        "level": 7,
+        "level": 0,
```
kay0u:
```suggestion
        "level": 7,
```

apps.json
```diff
@@ -985,7 +985,7 @@
     },
     "mantis": {
         "branch": "master",
-        "level": 1,
+        "level": 0,
```
kay0u:
```suggestion
        "level": 1,
```

PR #858, opened 2019-12-18
Improve + python3 support

kay0u:


list_builder.py
```diff
@@ -183,40 +183,38 @@ def get_zlib(url, verify=True):
...
-        app_rev = subprocess.check_output(["git", "ls-remote", app_url, "refs/heads/"+app_branch]).split()[0]
+        app_rev = subprocess.check_output(["git", "ls-remote", app_url, "refs/heads/"+app_branch]).split()[0].decode('utf-8')
         if not re.match(r"^[0-9a-f]+$", app_rev):
             error("Revision for %s did not match expected regex" % app)
             continue
 
-        if previous_rev is None:
-            previous_rev = 'HEAD'
+    if previous_rev is None:
+        previous_rev = app_rev
```
alexAubin: This PR is getting old, sorry, but as far as I remember, the whole handling of this "HEAD" case is kinda touchy. Do you remember why you think it's legit to change this to `app_rev` here ?

(Anyway the other changes looks okay, cheers :+1: )

kay0u: If I remember well, when we build for the first time an app which doesn't point to `HEAD` on github, we will look at the diff here:

https://github.com/YunoHost/apps/blob/dc722c2f6afb283b966528e05e42eca16e55081f/list_builder.py#L198

BUT, as the app doesn't point to `HEAD`, the `previous_rev` may be after the `app_rev`... With my solution, we don't go to this part of code for the first build.

kay0u: Since I added the `already_built_file` control to know if I can skip the build here:
https://github.com/YunoHost/apps/pull/858/files#diff-59862d9bc6c7dd67b153f8c75e042937R217
It's maybe not necessary. But at least it's easier to read like that (especially the optimize part) IMO of course!

alexAubin: Alrighty, yolo!

PR #861, opened 2020-01-02
Remove apps that install php version higher than 7.0

maniackcrudelis:
Without using a correct version of the helper.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1262,7 +1262,7 @@
         "category": "productivity_and_management",
         "level": 4,
         "revision": "HEAD",
-        "state": "working",
+        "state": "inprogress",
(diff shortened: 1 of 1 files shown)


yalh76:
Well, pixelfed use the last version of ynh_install_php__3, it's not good ?

maniackcrudelis:
I miss read the helper you were using. Sorry

yalh76:
> 
> 
> I miss read the helper you were using. Sorry

No problem, it's because it's integrated in _common.sh. I use to keep the original helper file on it's own and it's easier to check ^^ but not for pixelfed ^^

Thanks for your work ;)

PR #886, opened 2020-03-20
Cleaning / reorganizing the repo ?

alexAubin:
We're starting to have quite a lot of files in this repo and some of them are outdated or should be updated or reorganized... so here's an attempt

So far : 
- [x] `should_i_rebuild` ain't used anymore, replaced it with another script closer to what we currently do in the cron job (the idea being that now the cron job should just be calling this new script)
- [x] Having to feed the github credentials "by hand" in `list_builder.py` was madness, let's just store them in an unversioned file and read them from it when we need them.
- [x] There's no dev.json anymore, so no need for this whole `locales-dev` folder

Major remaining questions : 
- @Psycojoker : during the process in which we merged official and community.json to apps.json, I didn't know how to handle the whole stuff that so far happened in `should_i_rebuild.sh` ... Also the current folders `locales-official/` and `locales-community/` should probably somehow be merged but it's quite fuzzy how / where that stuff should happen...
- I don't know what to do with the `app_levels/` folder, as far as I know, it's only used by the app CI and it's overkill to try to translate these as it's never used for "actual users"
- Ideally it'd be nice to regroup all the small tooling in a dedicated folder but idk maybe that's me and my obsession with putting things in boxes lulz é_è
- I should probably import the "maintenance ping" tool in this repo (currently it's in tartiflette)
- We could really simplify the format of `apps.json` considering there's literally no app with `branch` set to something else that `master`, and just assume to use `HEAD` if there's no `revision` key set ...

Pretty minor stuff: 
- In the long term, imho it would be nice to switch to yaml for apps.json but meh i guess it would break many tools relying on this file
- Some of these scripts seems unused ? (idunno) Like `add_or_update.py` ?
- Spotted a the other day coming from `get_packager_email_addresses.sh` when an apps declared some `previous_maintainers`
- `check_id_unicity.py` could be turned into a `list_linter.py` where we check a few things (for example existence of the category key for each working app, ...) that we can add to travis
- In `list_builder.py`, I don't think there's such a need to have arguments handling so i'd vote to remove the whole argparse stuff


alexAubin:
So, in [this commit](https://github.com/YunoHost/apps/blob/90cfa0fe594a08fddf471910764ae5f101dff3e0/list_builder.py) I'm proposing a rewrite of list_builder.py

The current list_builder.py has many inconveniences such as: 
- We have so super-tricky logic that depends on the previously-built json file. This is super weird as we update the output json *only* if there are some specific changes. As a consequences, I've found that some changes (e.g. repo urls) have not been properly propagated on the output...
- We rely on data fetched from API (diff, commit date, ..) which depends on the type of forge the code is hosted on. And having to poke APIs means that things are slow as hell, you have to manage tokens, and so on...
- I think we don't try/catch all the work we do on each app ... so if the code fail for one app, it fails for the entire app

Instead, I propose to:
- keep an updated local git clone of every repo in the list (only the master branch with some ~50ish commits)
- analyze the commit history:
   - if using HEAD, find the most recent commit that affected a relevant file (manifest.json, scripts/, etc..)
   - if using a specific revision, just check that this revision is in the history
- build the dict per app with all relevant info ...

With this new builder we have:
- no super weird logic depending on previously-built file
- it's super quick to run (apart from initial git clone once) and independent of the type of forge it's hosted on. So we can imagine re-building every hours or so.
- (unrelated to building the list strictly speaking) because we have a git clone of everything, we can easily have tests like kayou was proposing (though not via travis) or compute statistics and other analysis like date-of-last-meaningful-update or list-all-apps-using-helper-X (for example I did this a lot when investigating edge cases for the permission system and other things)

PR #901, opened 2020-04-01
Update es.json

kay0u:
https://github.com/YunoHost/apps/blob/84aa2231d070f7e608b29fa715b741482b3a16dd/locales/en.json#L1214


alexAubin:
Uh that's probably why the tests fail since yesterday i guess

PR #880, opened 2020-03-18
End Official apps status

maniackcrudelis:
Following the announce on the forum, https://forum.yunohost.org/t/about-the-end-of-official-apps-in-yunohost/10530

Let's start the process of removing high_quality on all former Official apps.

We'll keep that PR opened to let packagers make their request for the tag on this branch.
So we'll have a trace of the requests for later.

- ~https://github.com/YunoHost-Apps/dokuwiki_ynh/issues/56~ > https://github.com/YunoHost/apps/pull/883
- ~https://github.com/YunoHost-Apps/etherpad_mypads_ynh/issues/85~ > https://github.com/YunoHost/apps/pull/884
- https://github.com/YunoHost-Apps/piwigo_ynh/issues/48
- https://github.com/YunoHost-Apps/opensondage_ynh/issues/60
- https://github.com/YunoHost-Apps/strut_ynh/issues/20
- https://github.com/YunoHost-Apps/synapse_ynh/issues/183
- ~https://github.com/YunoHost-Apps/wordpress_ynh/issues/88~ > https://github.com/YunoHost/apps/pull/881
- ~https://github.com/YunoHost-Apps/wallabag2_ynh/issues/75~
---
- Hextris > https://github.com/YunoHost/apps/pull/896
- My_webapp > https://github.com/YunoHost/apps/pull/900

PR #905, opened 2020-04-13
Ampache is broken

Aeris1One:
Ampache is broken and is maintainer-less so I removed high quality tag and made it notworking

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -99,6 +99,7 @@
     "ampache": {
         "branch": "master",
         "category": "multimedia",
+        "high_quality": true,
         "level": 7,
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Just tried to install, the app is indeed broken...

Aeris1One:
Why keeping high_quality ? One requirement for level 8/high quality tag is to have app maintained.

maniackcrudelis:
First because anyway, the app won't have the tag. It doesn't change anything.
And second, because there's already that PR #880 which remove the tag, to remove here will create a conflict.

PR #906, opened 2020-04-17
Dotclear2 HQ

kay0u:
# Validation template for High Quality tag request

Package URL: https://github.com/YunoHost-Apps/dotclear2_ynh

This template is designed to be used as it is by Apps group to validate requests from packagers for the tag High Quality.

Mandatory check boxes:
- [ ] The package is level 7.
- [ ] The package has been level 7 for at least 1 month.
- [ ] The package has been in the list for at least 2 months.
- [ ] The package is up to date regarding the packaging recommendations and helpers.
- [ ] The repository has a testing branch.
- [ ] All commits are made in testing branch before being merged into master.
- [ ] The list points to HEAD, not to a specific commit.
- [ ] The repository has a [`pull_request_template.md`](https://github.com/YunoHost/apps/blob/master/pull_request_template-HQ-apps.md)
- [ ] The package shows the YunoHost tile `yunohost_panel.conf.inc`

Optional check boxes:
- [ ] The package is level 7 for ARM as well.
*If the app is really important for the community, we can accept it with a broken ARM support. But this should be clearly explained and managed.*
- [ ] The app is up to date with the upstream version.  
*If this is possible with the last YunoHost version.*
- [ ] The package supports LDAP  
*If the app upstream supports it*
- [ ] The package supports HTTP authentication  
*If the app upstream supports it*

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -588,6 +588,7 @@
     "dotclear2": {
         "branch": "master",
         "category": "publishing",
+        "high_quality": true,
         "level": 7,
         "maintained": true,
         "revision": "HEAD",


maniackcrudelis:
I thought "Dotclear 2 HQ" was the name of a new app...

maniackcrudelis:
- What about the password here ? https://github.com/YunoHost-Apps/dotclear2_ynh/blob/testing/scripts/install#L156
It's not stored afterward. What is its function ?
- https://github.com/YunoHost-Apps/dotclear2_ynh/blob/testing/scripts/install#L174
Does the app has one or not ?
- No fail2ban ?
- Does the app really need a full chown $app ?
- A little bit of clean on the readme, nothing really important.

Would love to the CI result on your last commit, but it's far away into the waiting line... I'll run a local test.

kay0u:
>     * What about the password here ? https://github.com/YunoHost-Apps/dotclear2_ynh/blob/testing/scripts/install#L156
>       It's not stored afterward. What is its function ?

This password is only used to configure the blog, afterwards it is this code that manages the connection with ldap: https://github.com/YunoHost-Apps/dotclear2_ynh/blob/master/conf/class.auth.ldap.php (we cannot use this password anymore)
(The login method is defined [here](https://github.com/YunoHost-Apps/dotclear2_ynh/blob/6d38a59dab19b4f82fa1b661551b5b0688003c9e/scripts/install#L129-L133)

>     * https://github.com/YunoHost-Apps/dotclear2_ynh/blob/testing/scripts/install#L174
>       Does the app has one or not ?

There is an admin panel, and the username given at the installation is granted as super user.

>     * No fail2ban ?

Working on it (not sure if it's a possible thought).

>     * Does the app really need a full chown $app ?

Done :-)
 
>     * A little bit of clean on the readme, nothing really important.


Working on it

> Would love to the CI result on your last commit, but it's far away into the waiting line... I'll run a local test.

This is the result of the [last PR](https://github.com/YunoHost-Apps/dotclear2_ynh/pull/19): https://ci-apps-dev.yunohost.org/jenkins/job/dotclear2_ynh%20PR19%20(Kay0u)/5/console

maniackcrudelis:
The code looks good.

But to have 2 entry into the portal for the app isn't really welcome. For me at least. Hopefully I have only one installed.
A link on the interface of dotclear2 would be better.

And actually, that admin part gives me a blank page after login.

kay0u:
> But to have 2 entry into the portal for the app isn't really welcome. For me at least. Hopefully I have only one installed.

That's the point of YunoHost 3.7 and permissions system.

Only user in the permission `dotclear2.admin` see this second tile, (and can access to the admin part)

> And actually, that admin part gives me a blank page after login.

I have no problem on my side, let's fix that on IRC

maniackcrudelis:
Still trying (to use) the app.

Your readme says
```
Are LDAP and HTTP auth supported? Yes
Can the app be used by multiple users? Yes
```
So ldap indeed, but HTTP auth apparently not since I have some trouble with the authentication.
And multiple users ? You told me no, so how does it work ?

maniackcrudelis:
To summarize first
- A info about the users that can be only from YunoHost, just to explain.
- Also explain that if a password is smaller than 6 characters you gonna end up with a blank page with no more clues...

> That's the point of YunoHost 3.7 and permissions system.

Love this forced migration that bring more regressions and constraints than new features so far...

What bothers me is to have 2 tiles for the same app, just to have a link to the admin panel. Which should be accessible from the app itself. But looks like some dev love to hide the admin panel for their users... I'll never understand that...
A normal user will be lost not knowing where to look for. While a bot will go directly to the panel admin to try to brute force it... Useless...

Consider that I have 4 wordpress on my server, with such a rule for the admin panel, I would end up with 8 tiles !
Rather prefer a link at the bottom or somewhere else on the main page of the app.

kay0u:
>     * A info about the users that can be only from YunoHost, just to explain.

:+1: 

>     * Also explain that if a password is smaller than 6 characters you gonna end up with a blank page with no more clues...

> I've found a workaround: https://github.com/YunoHost-Apps/dotclear2_ynh/pull/23


> > That's the point of YunoHost 3.7 and permissions system.
> 
> Love this forced migration that bring more regressions and constraints than new features so far...

I don't feel like it's a regression, but I'm biased since I worked on it :-x

> What bothers me is to have 2 tiles for the same app, just to have a link to the admin panel. Which should be accessible from the app itself. But looks like some dev love to hide the admin panel for their users... I'll never understand that...
> A normal user will be lost not knowing where to look for. While a bot will go directly to the panel admin to try to brute force it... Useless...
> 
> Consider that I have 4 wordpress on my server, with such a rule for the admin panel, I would end up with 8 tiles !
> Rather prefer a link at the bottom or somewhere else on the main page of the app.

We've heard from you, and Joshua is already working hard on improving the permission system to fix the current system and provide more functionality.

maniackcrudelis:
> working hard on improving the permission system to fix the current system

Honestly that's exactly where the problem is.
I have absolutely no doubt that in a few months we're all going to say that the permission system is wonderful and far better than what we had before.
But, when you change things that way, you don't drop the current system which is perfectly working to replace it by a new one not completely tested and (it was obvious) not fully working.

So glad to hear that things will be better later in the future. But what for now ? To impose 2 tiles for one app to every users doesn't sound like a good idea.
If all apps do the same, we're soon end up with a thick forest of tiles, for only a few apps...*

> I've found a workaround: YunoHost-Apps/dotclear2_ynh#23

I wasn't expecting that much, but that's for the best.

maniackcrudelis:
All good :+1:.
Except that situation with the second tile for the admin.
Either this should be an option, or to be removed entirely, but to pop 2 tiles without a way to have only one is not a good solution.
Especially to have a link to the admin panel, which is a part of the app itself.

As said, if all apps start to show a second tile for their admin part, it's quickly going to be a mess with hundreds of tiles.
To have it as an option sounds like the best way to deal with this situation.

PR #910, opened 2020-04-21
Add mautrix_whatsapp

Gredin67:
Matrix-WhatsApp bridge. "mautrix-whatsapp" is the name of the upstream app https://github.com/tulir/mautrix-whatsapp

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1518,7 +1518,7 @@
         "subtags": [
             "chat"
         ],
-        "url": "https://github.com/YunoHost-Apps/mautrix-whatsapp_ynh"
+        "url": "https://github.com/YunoHost-Apps/mautrix_whatsapp_ynh"
     },
     "mediadrop": {
         "branch": "master",
--- a/apps.json
+++ b/apps.json
@@ -1510,6 +1510,16 @@
         ],
         "url": "https://github.com/YunoHost-Apps/mattermost_ynh"
     },
+    "mautrix_whatsapp": {
+        "branch": "master",
+        "category": "communication",
+        "revision": "HEAD",
+        "state": "working",
+        "subtags": [
+            "chat"
+        ],
+        "url": "https://github.com/YunoHost-Apps/mautrix-whatsapp_ynh"
+    },
     "mediadrop": {
         "branch": "master",
         "category": "multimedia",


Gredin67:
The app depends on synapse. Some discussions about that
yalh76
for CI test, maybe you can use somthing around https://github.com/YunoHost-Apps/writefreely_ynh/blob/45584b220e911c573643cb1e213e75fc42547818/scripts/install#L40-L42 that make do thing only during package check....
kayou
we use things like that to perform a pre-install: https://github.com/YunoHost-Apps/agendav_ynh/blob/dafa3e4230933be46b4f34a0ecd86d594c98b3ce/check_process#L7-L10

apps.json
```diff
@@ -1510,6 +1510,16 @@
         ],
         "url": "https://github.com/YunoHost-Apps/mattermost_ynh"
     },
+    "mautrix_whatsapp": {
+        "branch": "master",
+        "category": "communication",
+        "revision": "HEAD",
```
kay0u, line 1516: Please if you set `HEAD` create and work on the branch `testing`

alexAubin:
@Gredin67 : hey there ! Thank you for you work !

It helps if you can try to have a consistent ID for your app. Currently there is : 
- in the manifest, id is `mautrix_whatsapp`
- name of the repo is `mautrix-whatsapp`
- key in this PR is `mautrix_whatsapp`

so let's either keep and propagate the id with `_` or `-`

apps.json
Gredin67: Yep I'll do it just after merging my current PR in master. Then I'll do the next PRs on testing first

Gredin67:
1) Well I guess propagating ``_`` simplifies a lot, as the ``$app`` name is also used for postgres users and db. 
2) I was also thinking about simply calling the whole ``whatsapp``, as I'm not expecting other whatsapp related app to come. 
3) Or ``mwhatsapp``. Like [m]user naming scheme in e.g. IRC bridged room.
4) Last idea is ``mautrixwa``, knowing that other bridges exist for facebook messenger ``mautrixfb``, telegram ``mautrixtg``, and so on.

@tituspijean @Thatoo , what would you vote for? I think I would vote for 3).

Thatoo:
I like the idea of mautrixwa & co because it highlights Matrix and kind of weakened whatsapp and so on but I'm not sure it will be easy to distinguish one from the others so I guess mautrix_whatsapp is fine.'
maybe 
mautrix_wa
mautrix_fa
mautrix_tm
mautrix_tw
that could work, it's readable.
I don't like mwhatsapp.

apps.json
Gredin67: working version merged and testing branch created.

Gredin67:
Finally, I propagated mautrix_whatsapp everywhere. I just don't know how to change the github url to 
https://github.com/YunoHost-Apps/mautrix_whatsapp_ynh

alexAubin:
@Gredin67 : cheers ! Changing the name of the repo happens by going in the ['Settings' screen](https://github.com/YunoHost-Apps/mautrix_whatsapp_ynh/settings), and you can change the name

Took the liberty to do it

Note that github automatically redirects the old url https://github.com/YunoHost-Apps/mautrix-whatsapp_ynh to the new one (also when you do git push etc., it will just warn you that you should update it ideally, but it at least works for a couple of years)

Gredin67:
Hi, the master branch has been tested by 2 people now. Install and remove work from web and cli. Am I missing something for this PR to be validated?

PR #916, opened 2020-04-27
Add unlisted apps

maniackcrudelis:
A lot of old unlisted apps.

Added only apps :
- that had their last commit more than a 100 days ago.
- are not non-free
- are actually real apps, not a simple template


maniackcrudelis:
Why not, at least we would know where to put them.

PR #931, opened 2020-05-21
Switch to master's HEAD for diaspora

autra:


apps.json
```diff
@@ -469,7 +469,7 @@
         "branch": "master",
         "category": "social_media",
         "level": 3,
-        "revision": "b01c695cb81df24be2baa1fbe77f250c7f3195bd",
+        "revision": "0.7.13.0-ynh2",
```
alexAubin, line 472: Uuuuuuuh I don't know if that's correctly handled by list_builder ? But let's see how that goes I guess ¯\\\_(ツ)\_/¯

autra: Some other packages uses branch name. You expect tag to behave differently? I can make some tests if you want.

autra: Actually, what's the good practice here? Put master and be done with it? Or put tags, to be changed at each release, like I'm doing here?

autra: > [Applist builder error] Processing diaspora failed: Revision ain't in history ? 0.7.13.0-ynh2

:-O

autra:
Ok I'm doing like the others :-) PR updated.

apps.json
kay0u: We're expecting a commit hash or `HEAD` here.

alexAubin: Yeah we could support tags somehow I suppose, but the idea is that we switched to HEAD because people legitimately kept forgetting about updating the hash here ... so having a tag would be the same story. Using just HEAD you manage the release process of your apps independently of the list.

autra: Good for me!

PR #938, opened 2020-06-06
Update app's level

kay0u:


apps.json
```diff
@@ -1043,7 +1044,7 @@
     "gotify": {
         "branch": "master",
         "category": "dev",
-        "level": 7,
+        "level": 3,
```
yalh76, line 1047: Can be moved back to level 7 => https://ci-apps.yunohost.org/ci/job/784

kay0u, line 1047:
```suggestion
        "level": 7,
```

PR #939, opened 2020-06-09
Draft for an autopatch 'framework'

alexAubin:
Follow-up of https://github.com/YunoHost/issues/issues/1627 and discussions about the possibility to patch apps massively and automatically

So here's a draft of script that git clones all apps with level >= 4, apply a patch, and create a corresponding PR on all repos.

Usage looks like this : 

```
# Init local git clone for all apps
./autopatch --build-cache

# Apply patch in all local clones
./autopatch --apply explicit-php-version-in-deps

# Inspect diff for all apps
./autopatch --diff

# Push and create pull requests on all apps with non-empty diff
./autopatch --push explicit-php-version-in-deps
```

Resulting PR looks like this : https://github.com/YunoHost-Apps/piwigo_ynh/pull/56

### DIFF
--- a/autopatches/patches/explicit-php-version-in-deps/patch.sh
+++ b/autopatches/patches/explicit-php-version-in-deps/patch.sh
@@ -1,0 +1,12 @@
+PHP_DEPS="php-bcmath php-cli php-curl php-dev php-gd php-gmp php-imap php-intl php-json php-ldap php-mbstring php-mysql php-soap php-sqlite3 php-tidy php-xml php-xmlrpc php-zip php-dom php-opcache php-xsl php-apcu php-geoip php-imagick php-memcached php-redis php-ssh2 php-common"
+
+grep -q -nr "php-" scripts/* || exit 0
+
+for DEP in $PHP_DEPS
+do
+    NEWDEP=$(echo $DEP | sed 's/php-/php7.0-/g')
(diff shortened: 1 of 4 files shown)


autopatches/patches/explicit-php-version-in-deps/patch.sh
```diff
@@ -0,0 +1,12 @@
+PHP_DEPS="php-bcmath php-cli php-curl php-dev php-gd php-gmp php-imap php-intl php-json php-ldap php-mbstring php-mysql php-soap php-sqlite3 php-tidy php-xml php-xmlrpc php-zip php-dom php-opcache php-xsl php-apcu php-geoip php-imagick php-memcached php-redis php-ssh2 php-common"
+
+grep -q -nr "php-" scripts/* || exit 0
+
+for DEP in $PHP_DEPS
+do
+    NEWDEP=$(echo $DEP | sed 's/php-/php7.0-/g')
```
kay0u, line 7:
```suggestion
    NEWDEP=$(echo $DEP | sed 's/php-/php${YNH_DEFAULT_PHP_VERSION}-/g')
```

kay0u: We must change the requirements to 3.8 if we want this suggestion.

alexAubin:
Superseded by https://github.com/YunoHost/apps/pull/1080

PR #945, opened 2020-06-18
Update Funkwhale maintenance status and category

Sporiff:
Hi there!

This is just a quick change to reflect the current maintenance status/category of the Funkwhale package.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1,4 +1,4 @@
- {
+{
     "20euros": {
         "branch": "master",
         "category": "games",
--- a/apps.json
+++ b/apps.json
@@ -1,4 +1,4 @@
-{
+ {
     "20euros": {
         "branch": "master",
         "category": "games",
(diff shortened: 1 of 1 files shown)


apps.json
```diff
@@ -857,9 +857,9 @@
     },
     "funkwhale": {
         "branch": "master",
-        "category": "social_media",
+        "category": "multimedia",
```
alexAubin, line 860: Zblerg behind this there's a more general question of how do we categorize application that are activitity-pub based but whose social-networks capabilities are not the main focus (I think ?) ... for example Plume if I'm not mistaking should be in the publication / blog category. But clearly I understand that people would expect to find funkwhale in the multimedia category somehow.

Sporiff: Yeah, I would say Funkwhale falls pretty squarely into multimedia. We do support social elements through ActivityPub, but I'd say that was secondary to the media playback element. I was prompted to make this change by [this comment in the forum](https://forum.yunohost.org/t/taking-over-maintenance-of-an-application/11486/9), which highlights to me that the multimedia category is where people will look at first.

PR #958, opened 2020-07-26
Add zusam to apps.json

nrobinaubertin:


nrobinaubertin:
@alexAubin, can you merge ? I would like to have the results of the CI

yalh76:
> @alexAubin, can you merge ? I would like to have the results of the CI

As the app is in progress state, CI won't package_check it

nrobinaubertin:
Thanks for the answer.
The app is usable from my testing point of vue but I don't want to give false hope for a unknown user of yunohost. If I understand correctly, I should specify it as working and the decision lies on the user with the knowledge of the app level ?

alexAubin:
Ooooor we can add a test on the dev CI and see how that goes ;P

I can do it but you'll probably have to iterate a few times (fix stuff, re-add test) so it's better if you come to the app chatroom and add for an access by providing an ssh key

PR #977, opened 2020-09-26
Update app's level

yalh76:
- [x] armadietto https://ci-apps.yunohost.org/ci/job/1902
- [ ] garradin https://ci-apps.yunohost.org/ci/job/1914
- [x] jappix https://ci-apps.yunohost.org/ci/job/2088
- [x] snipeit https://ci-apps.yunohost.org/ci/job/1867

PR #978, opened 2020-09-28
Jappix level 4

alexAubin:
Ugh sorry for jumping late after https://github.com/YunoHost/apps/pull/977 but jappix is still supposed to be level 4 according to https://ci-apps.yunohost.org/ci/job/2088

(But i noticed a few times the CI reporting level 7 even though it clearly failed package linter ..)


kay0u (approved):
hmm, yeah, sorry

PR #1003, opened 2020-11-02
Modified squid3 as working app

anmol26s:
It would be good to test with CI before merging this.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -2882,7 +2882,7 @@
         "branch": "master",
         "category": "small_utilities",
         "revision": "HEAD",
-        "state": "notworking",
+        "state": "working",
         "url": "https://github.com/YunoHost-Apps/squid3_ynh"
     },
     "ssh_chroot_dir": {


ericgaspar:
well it's failing on Manifest...

`✘ Providing a LICENSE is mandatory`

alexAubin:
Indeed and there's a bunch of warnings in package_linter that would be nice to fix (my results are from the new / not yet merged version of package_linter) though the app may be installable etc.

```
 [Manifest]

! Setting the license as 'free' implies to write something about it in the README.md. Alternatively, consider using one of the codes available on https://spdx.org/licenses/ 
! The type 'port' for argument 'squid_port' is not recognized... it probably doesn't behave as you expect ? Choose among those instead : domain, path, boolean, app, password, user, string, display_text 

 [_Common.Sh Script]


 [Install Script]

! You should not use `ynh_package_install` or `apt-get install`, use `ynh_install_app_dependencies` instead 

 [Remove Script]

! You probably don't need to run 'yunohost app ssowatconf' in the app self. It's supposed to be ran automatically after the script. 

 [Upgrade Script]

! You should not use `ynh_package_install` or `apt-get install`, use `ynh_install_app_dependencies` instead 
! Using ynh_script_progression --time should only be for calibrating the weight (c.f. --weight). It's not meant to be kept for production versions. 

 [Backup Script]


 [Restore Script]

! You should not use `ynh_package_install` or `apt-get install`, use `ynh_install_app_dependencies` instead 

 [Misc Helper Usage / Consistency]

! ynh_delete_file_checksum is now an official helper since version '3.3.1' 
! Using official helper ynh_exec_warn_less implies requiring at least version 3.2.0, but manifest only requires 2.7.14 
! Using official helper ynh_handle_getopts_args implies requiring at least version 3.2.2, but manifest only requires 2.7.14 
! Using official helper ynh_print_info implies requiring at least version 3.2.0, but manifest only requires 2.7.14 
! Using official helper ynh_script_progression implies requiring at least version 3.5.0, but manifest only requires 2.7.14 
! Using official helper ynh_systemd_action implies requiring at least version 3.5.0, but manifest only requires 2.7.14 
! The install script expose a port on the outside with 'yunohost firewall allow' but doesn't use 'yunohost service add' with --needs_exposed_ports ... If your are ABSOLUTELY SURE that the service needs to be exposed on THE OUTSIDE, then add --needs_exposed_ports to 'yunohost service add' with the relevant port number. Otherwise, opening the port leads to a significant security risk and you should keep the damn port closed ! 
✘ Providing LICENSE is mandatory 
```

anmol26s:
This commit should fix things.
https://github.com/YunoHost-Apps/squid3_ynh/commit/44eb95620bbaaaf37eab93ef041f2cff50744fb0

alexAubin:
Weell the tests results ain't really great .. it didnt run any tests : 

https://ci-apps-dev.yunohost.org/jenkins/job/squid3_ynh%20(aleks)/lastBuild/console

I believe it's quite probably related to the fact that there are 0's all over the place in https://github.com/YunoHost-Apps/squid3_ynh/blob/master/check_process#L10-L23 ? :/

anmol26s:
I have modified restore and backup to true as in mumble app. This should check backup and restore. As it do not use web routes other checks would not be required.

alexAubin:
Alright I now realized that this app is not a webapp so maybe what's missing is that you need to enable "setup_nourl" inside check_process ... though I'm no expert regarding package_check behaviors ...

Anyway, after modifying this I retriggered a job here : https://ci-apps-dev.yunohost.org/jenkins/job/squid3_ynh%20(aleks)/

we'll see

alexAubin:
@anmol26s : results are available (only restore seems to be failing)

https://ci-apps-dev.yunohost.org/jenkins/job/squid3_ynh%20(aleks)/2/console

Note that the package linter seems to suggest relevant changes to fix the issue

anmol26s:
I think @ericgaspar hasdone the required changes. 
Thanks @ericgaspar

kay0u:
I think we could merge this PR, and you can continue debugging it with the CI

alexAubin (approved):

PR #1011, opened 2020-11-06
Set revision for couchpotato, owntracks, phpldapadmin and wetty

alexAubin:
Some of these have quite a lot of significant commits since the commit flagged here ...

Really tempted to report using HEAD as an error in linter because having to check for this manually is just a waste of time

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -449,7 +449,7 @@
         "category": "multimedia",
         "level": 7,
         "maintained": false,
-        "revision": "2fbe760a038a1e65f574d30ffb293c7499229235",
+        "revision": "HEAD",
         "state": "working",
         "subtags": [
             "download"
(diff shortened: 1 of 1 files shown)


ericgaspar (approved):

kay0u (approved):

kay0u:
> Really tempted to report using HEAD as an error in linter because having to check for this manually is just a waste of time

I don't know, if a packager commits without testing its code on master, users can have a lot of surprises during an upgrade

alexAubin:
Agreed but : 
- nowadays 95+% of (working) apps are using HEAD
- I believe all of us easily forget that "ah yeah maybe this app ain't using HEAD"
- ... which on the contrary create situations where some relevant and important commits were pushed to the branch, but ain't available to users because a specific revision is set in apps.json
- (not sure about that but I believe the CI always tests HEAD ? so the level in apps.json may not reflect the level of the actual commit)
- with the upcoming 4.1, upgrades will only be made available if there's a version changes

Overall yeah, we kind of *have* to trust people to not push random untested stuff to the master branch :/ Imho the right thing to do to address this would be to have a way to wait for a result from the CI to validate a given commit before propagating the commit to the "built" list ... (I think we talked briefly about this before but this is just quite a lot of work)

yalh76 (approved):

PR #1013, opened 2020-11-06
Removing caliopen

yalh76:
As creator and maintainer of this unfinished package I prefer to remove it and plan to delete the corresponding repository


alexAubin (approved):
Wokay I guess ?

PR #1023, opened 2020-11-22
Update high quality checklist (c.f. new definitions for level 7, 8, 9)

alexAubin:
c.f. the ongoing discussion to rework the definition of level 7, 8 and 9

- I removed the items that should now be handled by level 7 and 8 from the "high quality" definition
- I reworked some wording / formatting  for the existing criterias
- I added some criterias which I think should be relevant nowadays, such as fail2ban integration (if relevant) and explicitly encouraging the reviewers to look for any obvious security issue or borderline practice (during the initial validation process).
- I removed the duplicate explanation inside the README about the exact criterias and instead pointed to the hq_validation_template.md, because it's hell to maintain both things

### DIFF
--- a/hq_validation_template.md
+++ b/hq_validation_template.md
@@ -1,26 +1,22 @@
 # Validation template for High Quality tag request
 
-Package URL: 
+Package URL:
 
-This template is designed to be used as it is by Apps group to validate requests from packagers for the tag High Quality.
+This template is designed to be used by the Apps group to validate requests from packagers for the tag High Quality.
 
-Mandatory check boxes:
-- [ ] The package is level 7.
-- [ ] The package has been level 7 for at least 1 month.
-- [ ] The package has been in the list for at least 2 months.
-- [ ] The package is up to date regarding the packaging recommendations and helpers.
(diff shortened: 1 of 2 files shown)


hq_validation_template.md
```diff
@@ -1,26 +1,22 @@
...
-Package URL: 
+Package URL:
 
-This template is designed to be used as it is by Apps group to validate requests from packagers for the tag High Quality.
+This template is designed to be used by the Apps group to validate requests from packagers for the tag High Quality.
 
-Mandatory check boxes:
-- [ ] The package is level 7.
-- [ ] The package has been level 7 for at least 1 month.
-- [ ] The package has been in the list for at least 2 months.
```
kay0u, line 10: Do we really want to drop this? It allows us to know if the packager will maintain its app seriously. (But at the same time, if he makes the effort to make the request, he is normally motivated to stay).

alexAubin: The idea is that basically this is now part of level 8 (c.f. the "long term quality" check)

PR #1027, opened 2020-11-26
Change domoticz_ynh to Yunohost-Apps

Krakinou:
As discuted with anubister who developped the previous version of the app, I packaged a new version that has all script working, do not require compiling and allow to update to the last upstream version.
According to my tests, moving from previous anubister version to current one is working


kay0u (approved):
Thanks!

PR #1036, opened 2020-12-12
Add django-for-runners

jedie:
Project Homepage: https://github.com/jedie/django-for-runners

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -551,6 +551,12 @@
         ],
         "url": "https://github.com/YunoHost-Apps/distbin_ynh"
     },
+    "django-for-runners": {
+        "branch": "master",
+        "revision": "HEAD",
+        "state": "notworking",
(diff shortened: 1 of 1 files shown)


tituspijean (approved):
🚀

apps.json
```diff
@@ -551,6 +551,12 @@
         ],
         "url": "https://github.com/YunoHost-Apps/distbin_ynh"
     },
+    "django-for-runners": {
+        "branch": "master",
```
yalh76, line 555: You should add a master branch adn make it the default branch

yalh76 (changes requested):

apps.json
jedie: Ah sorry. Created: https://github.com/YunoHost-Apps/django-for-runners_ynh/tree/master

kay0u (approved):

PR #1059, opened 2021-01-20
add PHP Sevrer Mon Apps

BenoitCier:
https://ci-apps-dev.yunohost.org/ci/job/60
Thanks for you help!

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -2401,6 +2401,12 @@
         ],
         "url": "https://github.com/YunoHost-apps/phpmyadmin_ynh"
     },
+    "phpservermon": {
+        "branch": "master",
+        "revision": "04c1ee06843ea4a7fd00b4eb6c381286f2562f1b",
+        "state": "working",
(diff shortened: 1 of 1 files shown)


alexAubin (approved):
Thank you for your work !

I might in fact install this app to replace my own home-made script :P https://github.com/alexAubin/my-little-minotor/

apps.json
```diff
@@ -2401,6 +2401,12 @@
         ],
         "url": "https://github.com/YunoHost-apps/phpmyadmin_ynh"
     },
+    "phpservermon": {
+        "branch": "master",
+        "revision": "04c1ee06843ea4a7fd00b4eb6c381286f2562f1b",
```
alexAubin, line 2406: Note that we recommend using HEAD to avoid having to come back here to update the revision everytime you publish an update of the package (which generally ends up in people forgetting about it)

ericgaspar (approved):

PR #1067, opened 2021-01-31
Add maintained pi-hole

benjaminwolkchen:


ericgaspar (commented):
Why create a second entry for Pi-Hole and not use the existing one?

alexAubin:
(Or even make a PR to the original repo ...)

PR #1069, opened 2021-02-05
Add Yusers application

ziima:
Add yusers to apps.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3631,6 +3631,16 @@
         ],
         "url": "https://github.com/YunoHost-Apps/yunomonitor_ynh"
     },
+    "yusers": {
+        "branch": "main",
+        "category": "system_tools",
+        "revision": "HEAD",
+        "state": "working",
+        "subtags": [
+            "others"
+        ],
+        "url": "https://gitlab.com/ziima/yusers"
+    },
     "z-push": {
         "branch": "master",
         "category": "synchronization",


apps.json
```diff
@@ -3631,6 +3631,16 @@
...
     },
+    "yusers": {
+        "branch": "main",
+        "category": "system_tools",
+        "revision": "HEAD",
+        "state": "working",
+        "subtags": [
+            "others"
+        ],
+        "url": "https://gitlab.com/ziima/yusers"
```
alexAubin, line 3642:
```suggestion
        "url": "https://gitlab.com/ziima/yusers_ynh"
```

Ideally the repository should really be named 'appid_ynh', otherwise that may cause various issues :sweat_smile:

ziima: The repository contains both, the application itself and the Yunohost integration. In such case I'd rather keep only the name of the application itself.

Aeris1One: Uh, gitlab URL are allowed ?

ziima: @Aeris1One I was able to install it using CLI with no problem.

Aeris1One: Well, that's a recent improvement but I'm not sure the webadmin will allow a gitlab repo, not sure though, let someone with better understanding of YunoCore than me answer

AshenColors: My CLI doesn't allow it, and the webadmin *definitely* doesn't allow it.

```
user@computer:~$ sudo yunohost app install https://gitlab.com/ziima/yusers
Error: https://gitlab.com/ziima/yusers is not a valid app url: app url are expected to look like https://domain.tld/path/to/repo_ynh
Error: Unknown app```

ziima: It worked for me at the time.

I was able to find that the change was caused by the https://github.com/YunoHost/yunohost/commit/171b3b25d9d4cb7aad8b2cbbc5aeb7eee6243a1f

alexAubin: I don't know what to say, it's just very much better that the repo should be called `foobar_ynh` ... There's just too many pieces of code that may assume that the repo is named this way. The fact that this worked before was a bit "random" (like the code was not formalized-enough).

Similarly having the upstream app being developed "inside" the app itself is not a standard practice either and it's kind of a mess because you end up with a `manifest.json`, a `scripts` folder and other files and folers that are specific to yunohost - but which are not reflected in their naming (eg they are not inside a `ynh` folder or whatever) so everything is mixed up between the app code and the packaging code... I guess it works, but it's just messy.

With that said, that last point should be explored because there are many situations where having a `ynh` folder (similarly to the `debian` folder in debian-packaged software) directly inside the upstream would be helpful and simplify maintenance... But there's some deeper thoughts to be given to this to design and implement it properly

ziima: > The fact that this worked before was a bit "random"

I'm not very happy about that, but I understand.

> I guess it works, but it's just messy.

Yes, it is. I tried to find a way around, but didn't find any.

> With that said, that last point should be explored because there are many situations where having a ynh folder (similarly to the debian folder in debian-packaged software) directly inside the upstream would be helpful and simplify maintenance... But there's some deeper thoughts to be given to this to design and implement it properly

This would be awesome :+1: For some apps, such as `yusers` which are Yunohost specific, doesn't really make sense to create two repositories. The app needs Yunohost to work at all. Ability to move the Yunohost packaging files into a subdirectory would be a great improvement for these apps.

PR #1077, opened 2021-02-19
Add pepettes

zamentur:


kay0u (approved):
> Money
Get away
You get a good job with good pay and you're okay
Money
It's a gas
Grab that cash with both hands and make a stash
New car, caviar, four star daydream
Think I'll buy me a football team

:headphones:

PR #1080, opened 2021-02-21
Autopatch for new permission system

alexAubin:
Follow-up of https://github.com/YunoHost/package_linter/pull/92 , there's probably still like 100+ apps to be migrated to the new permission system and this is gonna take a gazillion hours ... So let's try to automate it (at least partially)

This is a respawn / continuation of https://github.com/YunoHost/apps/pull/939

The proposed script magically fixes/adapt all the app scripts to migrate to the new permission system using a bunch of super obscure and arbitrary regexes...

As explained in the original PR, the best thing to do is probably to check the diff directly : 

```
# Init local git clone for all apps
./autopatch --build-cache

# Apply patch in all local clones
./autopatch --apply new-permission-system

# Inspect diff for all apps
./autopatch --diff

# (Once ready) Push and create pull requests on all apps with non-empty diff
./autopatch --push explicit-php-version-in-deps
```

I reviewed a bunch of diffs and it seems ~okay (modulo that it doesn't cover non-trivial cases of course... but still should save some volunteer time)


alexAubin:
Doing more tests, I realize quite a lot of apps (117 !) are apparently already using the new permission system :o 

Still the patch generates a non-zero diff for 60 apps : 

```
airsonic
bitwarden
bozon
cheky
civicrm_drupal7
concrete5
couchpotato
diagramsnet
diaspora
drupal
drupal7
emailpoubelle
etherpad_mypads
ffsync
freshrss
ghost
grafana
horde
hubzilla
jeedom
jenkins
keeweb
leed
libreto
lstu
lufi
mailman
mastodon
meilisearch
moodle
multi_webapp
my_webapp
nextcloud
onlyoffice
opensondage
peertube
peertube-search-index
petitesannonces
phpsysinfo
pleroma
plume
prettynoemiecms
radicale
redirect
rss-bridge
spftoolbox
spip
strut
syncthing
ttrss
tyto
wallabag2
weblate
wikijs
wordpress
writefreely
z-push
zabbix
zap
zerobin
```

autopatches/autopatch.py
```diff
@@ -0,0 +1,188 @@
...
+    if not isinstance(cmd, list):
+        cmd = cmd.split()
+    if in_folder:
+        cmd = ["-C", in_folder] + cmd
+    cmd = ["git"] + cmd
+    return subprocess.check_output(cmd, env=my_env).strip().decode("utf-8")
+
+
+# Progress bar helper, stolen from https://stackoverflow.com/a/34482761
+def progressbar(it, prefix="", size=60, file=sys.stdout):
```
kay0u, line 42: https://github.com/tqdm/tqdm is easy to use for this.

autopatches/autopatch.py
```diff
@@ -0,0 +1,188 @@
...
+
+def fork_if_needed(repo, s):
+
+    repo_name = repo.split("/")[-1]
+    r = s.get(github_api + f"/repos/{login}/{repo_name}")
+
+    if r.status_code == 200:
+        return
+
+    r = s.post(github_api + f"/repos/{repo}/forks")
```
kay0u, line 134: Fork may takes time, is it ok to create it and not check if the fork is done? (I guess yes as we don't do a lot of fork compare to the amount of apps, but I prefer to ask)

autopatches/patches/new-permission-system/patch.sh
```diff
@@ -0,0 +1,66 @@
...
+	ynh_legacy_permissions_delete_all
+
+	ynh_app_setting_delete --app=\$app --key=is_public
+fi'
+
+for SCRIPT in "remove upgrade backup restore change_url"
+do
+    [[ -e $SCRIPT ]] || continue
+
+    perl -p0e 's@.*ynh_app_setting_.*protected_.*@@g' -i $SCRIPT
```
kay0u, line 40: just out of curiosity, why `per`l and not `sed`?

kay0u (approved):
LGTM otherwise

autopatches/patches/new-permission-system/patch.sh
alexAubin: Ah yes good question haha ... Because the first thing I was working on was finding a way to do a multiline replace, which you can with perl, but not with sed (c.f. the other regex patching "if ...(.|\n) fi") ... but in fact most of these could be done with sed ...

autopatches/autopatch.py
alexAubin: Uuuuuuh yeah good question ? I don't really know, this part of the code has not really been tested (or maybe back in the days when i first wrote this). Naively I would think the server returns 200 only after effectively creating the fork ? But maybe not ...

autopatches/autopatch.py
alexAubin: Indeed that's interesting, though one advantage of having this function in the code is that we don't have to pip install anything :s

autopatches/patches/missing-seteu-in-change_url/patch.sh
```diff
@@ -0,0 +1,10 @@
+
+cd scripts/
+
+if [ ! -e change_url ] || grep -q 'ynh_abort_if_errors' change_url
+then
+    # The app doesn't has any change url script or already has ynh_abort_if_error
+    exit 0
+fi
+
+sed 's@\(source /usr/share/yunohost/helpers\)@\1\nynh_abort_if_errors@g' -i change_url
```
kay0u, line 10: we should probably add a backup/restore mechanism right before `ynh_abort_if_errors`

alexAubin: Zblerg yeah I saw that on several apps and in fact I don't understand why it's done this way ... Shouldn't we just apply set -eu as quick as possible at the beginning of the script ...?

(In fact there's a similar weird stuff in upgrade script were we patch "backward compatibility stuff" *before* the safety backup ... but imho the point of the safety backup is precisely to save the app state before any modification starts to be applied...)

ericgaspar (approved):

PR #1087, opened 2021-02-28
Add Komga 

navanchauhan:


apps.json
```diff
@@ -1556,6 +1556,16 @@
...
         "url": "https://github.com/YunoHost-Apps/kodi_ynh"
     },
+    "komga": {
+        "branch": "master",
+        "category": "reading",
+        "revision": "HEAD",
+        "state": "working",
+        "subtags": [
+            "comics"
+        ],
```
alexAubin, line 1566: That subtag doesn't exists though, so proposing to drop it (or maybe "books" but idk)

alexAubin (approved):
LGTM apart from small comment about subtag

apps.json
navanchauhan: hmm, sure books is alright for now

navanchauhan: changed it

ericgaspar (approved):

PR #1136, opened 2021-04-23
README generator

alexAubin:
So today I went berserk after seeing again many PRs just about updating the README.

Then I thought "hey let's make an autopatch for this".

Then I started to think that we should patch other things in the README (such as removing the "supported architecture" section which doesn't make sense because we only have tests for x86)

Then started to rework the entire README because there are too many sections, too many links all over the place, etc

Then realized that this would be hell to update once again the 200+ README of every app

Three tableflips later (╯°□°）╯︵ ┻━┻ , I decided that we just loose a crazy ass amount of time with this whole README shit, we should standardize it, code is law, etc.

So here we are: 
- I created a jinja template for READMEs
- The README is automatically filled with info from the manifest.json, including: 
   - description
   - version number
   - new "upstream" info ([c.f. example PR on example_ynh](https://github.com/YunoHost/example_ynh/pull/155)), containing the website, demo, etc
- It's also filled with screenshots, automatically using screenshots inside `doc/screenshots/`
- It's also filled with the content of `doc/DISCLAIMER.md`. This part is meant to contain all important info (need for full domain, is LDAP/SSO working, unsupported arch, amount of RAM required, how to finalize install, how to configure the app, ...)
    - some of these info such as RAM needed or unsupported archs should be further standardized, but let's see this later (sigh) ...
    - this app could / should be displayed during the app install process, c.f. the old idea of "displaying the README", but displaying the entire README is probably too much
- The idea is that this script should automatically be applied somehow. We should discuss how exactly. Maybe a bot being triggered each time an app receives commit on the `manifest.json` or stuff in `doc/`

Current usage is: 

`python3 ./make_readme.py /path/to/the/app`

Then that directly applies the change to the README.md inside the app folder.

Example output : https://github.com/YunoHost/example_ynh/blob/new-standardized-readme/README.md


tools/READMEs/make_readme.py
```diff
@@ -0,0 +1,40 @@
+from jinja2 import Environment, FileSystemLoader
+import json
+import os
+import sys
+
+if len(sys.argv) <= 1:
```
jedie, line 6: Use argparse

tools/READMEs/make_readme.py
```diff
@@ -0,0 +1,40 @@
...
+    raise Exception("You should provide the path to the app as first arg")
+
+app = sys.argv[1]
+
+if not os.path.exists(app):
+    raise Exception("App path provided doesn't exists ?!")
+
+env = Environment(loader=FileSystemLoader('./templates'))
+
+for lang, lang_suffix in [("en", ""), ("fr", "_fr")]:
```
jedie, line 16: Move it to a function and make it possible to import this file

tools/READMEs/templates/README.md.j2
```diff
@@ -0,0 +1,53 @@
...
+{% if upstream.website %}* Official app website: {{ upstream.website }}{% endif %}
+{% if upstream.userdoc %}* Official user documentation: {{ upstream.userdoc }}{% endif %}
+{% if upstream.admindoc %}* Official admin documentation: {{ upstream.admindoc }}{% endif %}
+{% if upstream.code %}* Upstream app code repository:  {{ upstream.code }}{% endif %}
+* YunoHost documentation for this app: https://yunohost.org/app_{{manifest.id}}
+* Report a bug: https://github.com/YunoHost-Apps/{{manifest.id}}_ynh/issues
+
+## Developer info
+
+Please send your pull request to the [testing branch](https://github.com/YunoHost-Apps/{{manifest.id}}_ynh/tree/testing).
```
jedie, line 44: Add blocks to all sections. So it's possible to overwrite everything

tools/READMEs/make_readme.py
alexAubin: Done in https://github.com/YunoHost/apps/pull/1136/commits/f8c150a217aa2d81134c97011d66d55c8f53fa06

tools/READMEs/make_readme.py
alexAubin: Done in https://github.com/YunoHost/apps/pull/1136/commits/231890575060a9b4d6a5c3a56978ad0f9c075617

tools/READMEs/templates/README.md.j2
alexAubin: Mokay but there's not really any design for this ... the current design doesn't allow the app to override the structure of the README, it's really about fetching info from manifest.json, and injecting the contents from doc/screenshots/ and doc/DISCLAIMER.md

PR #1159, opened 2021-05-21
Auto-update-README webhook

alexAubin:
Follow-up of https://github.com/YunoHost/apps/pull/1136

This adds a small webhook service to auto-commit README updates

c.f. an example here on this dummy PR, where I updated the manifest.json's `upstream` data, which automatically triggered a commit from the bot: https://github.com/YunoHost-Apps/helloworld_ynh/pull/3/commits

The not-so-cool thing though is that I had to promote the yunohost-bot account from Member to Admin of the organization, such that it has write accesses to all repos to be able to push commits...


nils-van-zuijlen:
Is it possible for it to check for the existence of an `upstream` key in the manifest? Or that the README is already a generated one?

I currently have a bugfix PR (https://github.com/YunoHost-Apps/humhub_ynh/pull/21) that's blocked because the bot keeps rewriting the readmes (they are not auto-generated).

alexAubin:
Yes, it *does* test if there's a upstream key: https://github.com/YunoHost/apps/blob/autoreadme-webhook/tools/README-generator/make_readme.py#L18

Your PR does define the new upstream key, which is why it autogenerates / autoupdate the README, because that's precisely the purpose of the mecanism ... I'm not sure to understand what you behavior you expect

nils-van-zuijlen:
Oops, seems like I failed my commit.

Still, I would prefer that the webhook checks for presence of the comment at the top of the readme itself, if possible.

alexAubin:
Mokay but the point is that there should not be any explicit "opt in" mecanism, otherwise the whole transition to the auto-generated READMEs is gonna be even more complex and slow whereas it's supposed to just be handled automatically because there should be no reason to not use the new standard ... or maybe there is, but then let's discuss those.

PR #1162, opened 2021-05-22
zerobin renamed privatebin

ericgaspar:
Zerobin_ynh as been renamed privatebin_ynh to avoid confusion.


yalh76:
Is upgrade from zerobin to privatebin been  handle ?

ericgaspar:
> Is upgrade from zerobin to privatebin been handle ?

No. (last time I tried to migrate an app it was so unsuccessful... (CodiMD -> Hedgedoc)

PR #1165, opened 2021-05-26
Auto-update-README

nils-van-zuijlen:
This PR adds the following features:

## Ease of running the make_readme script from a venv
Previously, once in a venv, you'd have to run `python make_readme.py` because the shebang didn't account for env.
Now you can use `./make_readme.py`

## Ability to run the script from any folder
Previously, the templates were found relative to the shell working directory.
Now they are found relative to the script file.

### DIFF
--- a/tools/README-generator/make_readme.py
+++ b/tools/README-generator/make_readme.py
@@ -3,6 +3,7 @@
 import argparse
 import json
 import os
+from pathlib import Path
 
 from jinja2 import Environment, FileSystemLoader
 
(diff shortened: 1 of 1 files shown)


tools/README-generator/make_readme.py
```diff
@@ -1,54 +1,65 @@
...
 
-def generate_READMEs(app_path):
+def generate_READMEs(app_path: Path):
 
-    if not os.path.exists(app_path):
+    if not app_path.exists():
         raise Exception("App path provided doesn't exists ?!")
 
-    manifest = json.load(open(os.path.join(app_path, "manifest.json")))
+    manifest = json.load(open(app_path / "manifest.json"))
```
alexAubin, line 16: omfg i thought we could only do this in the context of pytest

This is fucking awesome, I've been dreaming of this :star_struck:

alexAubin (approved):

PR #1180, opened 2021-06-13
Port readme generator webhooks from flask to sanic

Psycojoker:
Hello,

This should normally reduce greatly the memory/number of process needed impact (most of the code is coming from both webhooks and yunorunner) by having only one python process instead of 5 gunicorn runners.

I've tested as much as possible locally but this will still need a prod testing.

I'll see if I can do that tomorrow.

Kind regards,

### DIFF
--- a/tools/README-generator/requirements.txt
+++ b/tools/README-generator/requirements.txt
@@ -1,4 +1,2 @@
-argparse
 jinja2
-github-webhook==1.0.4
-gunicorn==20.1.0
+sanic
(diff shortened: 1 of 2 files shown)


Psycojoker:
So this is running in prod for now and it seems to be working so I guess we can merge in a few days if we like this perf improvement (this doesn't brings any new real features)

alexAubin:
Yolo then ¯\\\_(ツ)\_/¯

PR #1183, opened 2021-06-22
Add official websites

rosano:
Hi/Bonjour, I am compiling a list of self-hosted applications based on official websites and so I added them to `apps.json`. If it helps YunoHost, feel free to use it.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -6,7 +6,8 @@
         "maintained": true,
         "revision": "HEAD",
         "state": "working",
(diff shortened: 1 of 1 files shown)


tituspijean:
I feel that it may be redundant with the information available in each app manifest. I am not sure it is needed to display the app websites at the first level of the catalog.

rosano:
> Indeed, the info should already be in the built list : https://app.yunohost.org/default/v2/apps.json

Oh I didn't know about that one, thank you.

PR #1190, opened 2021-06-28
Weblate is (unofficialy) maintained

ericgaspar:
We should probably remove the `Not maintained` flag as it is maintained...


kay0u (approved):

yalh76:
but there is no maintainer in the manifest.json :'(

PR #1192, opened 2021-07-04
Don't overwrite translations from manifests + remove stale fr translations for descriptions

alexAubin:
c.f. discussion today on the chat : we have weird descriptions showing up in the catalog which are from the locales/fr.json and that's quite unexpected because there's already a fr string in the manifest.json that's probably more up-to-date


kay0u (approved):
👍

ericgaspar:
👍

ericgaspar (approved):

PR #1203, opened 2021-07-21
Jappix

ericgaspar:
Set Jappix as not working

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1578,7 +1578,7 @@
         "category": "communication",
         "maintained": false,
         "revision": "HEAD",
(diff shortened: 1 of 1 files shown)


alexAubin:
But didn't you somewhat fix it in https://github.com/YunoHost-Apps/jappix_ynh/pull/28 just a couple hours ago ? :thinking:

ericgaspar:
For some reason the install in my server gives me an error: `erreur : serveur indisponible`. As the project seems to be dead for years now... I don't know if it's worth the time spended 😅

PR #1241, opened 2021-09-18
Add n8n to catalog

ericgaspar:
* Add n8n to catalog

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -2009,7 +2009,7 @@
     },
     "n8n": {
         "antifeatures": [
-            "Not totally free upstream"
+            "not-totally-free"
         ],
         "category": "iot",
         "state": "working",
--- a/apps.json
+++ b/apps.json
@@ -2007,6 +2007,11 @@
         ],
         "url": "https://github.com/YunoHost-Apps/mytinytodo_ynh"
     },
+    "n8n": {
+        "category": "iot",
+        "state": "working",
+        "url": "https://github.com/YunoHost-Apps/n8n_ynh"
+    },
     "navidrome": {
         "category": "multimedia",
         "level": 8,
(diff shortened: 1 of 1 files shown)


tituspijean (approved):

fflorent:
Can you tell me what do you think of this issue?

https://github.com/YunoHost-Apps/n8n_ynh/issues/6 (*Not totally free license*)

If I quote [your documentation](https://yunohost.org/en/contribute/packaging_apps/quality_tests/yep):
> The application list apps.json only accept packages with a free license, as well as the license for the contained application. Some free applications require non-free dependencies (example: MP3, drivers, etc.). In this case, you should add &dep-non-free to the acronym and if possible give details in the README.md of the package, in this case the integration will be accepted on a case-by-case basis.

I don't know whether it would be compatible with the apps.json file here.

tituspijean:
That's a valid remark! I think we cannot allow such license in the catalog. 😞

Limezy:
>  😞 Maybe it is time to revive the non-free apps list.

@tituspijean I think I found the exact same problem for my Outline app... 
Its licence is "Business Source License 1.1" (https://github.com/outline/outline/blob/main/LICENSE)

Do you confirm that I can't add it to the app catalog ? 😢 😭

ericgaspar:
@tituspijean I propose to merge this with the addition of the licence antifeature

fflorent:
LGTM

nsakibp:
Please add

apps.json
```diff
@@ -2007,6 +2007,14 @@
         ],
         "url": "https://github.com/YunoHost-Apps/mytinytodo_ynh"
     },
+    "n8n": {
+        "antifeatures": [
+            "Not totally free upstream"
```
tituspijean, line 2012:
```suggestion
            "not-totally-free"
```

tituspijean (changes requested):

tituspijean (approved):

PR #1250, opened 2021-10-03
Update app's level

alexAubin:
- [x] [archivebox](https://ci-apps.yunohost.org/ci/job/3406)

Other regressions to level 6 are related to recent changes on the CI: 
- CI was not properly testing some usecase of $change_url, in which some apps have unbound $domain var (eg freshrss)
- Linter now reports legacy permissions as a warning
- CI now blocks level 7+ for apps with unsafe $final_path perms

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -106,7 +106,7 @@
     },
     "archivebox": {
         "category": "small_utilities",
-        "level": 1,
+        "level": 7,
         "state": "working",
         "subtags": [
             "archiving"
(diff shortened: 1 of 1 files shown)


tituspijean:
Should we re-run tests for the "regressed" apps?

alexAubin:
I havent checked all of those but from the quick look I had it was understood regressions from changes in the CI .. but feel free to have a deeper look (I mean apart from archivebox it's all about level 8->6 or 7->6)

kay0u (approved):
LGTM

PR #1251, opened 2021-10-08
Add ssbroom to apps.json

mhfowler:
Not all the tests are passing yet, but the app is working


yalh76:
Would be great to create a testing branch first. and it seems that your readme still contain default informations

PR #1262, opened 2021-11-03
Add uptime-kuma

ketsapiwiq:
Latest CI says:
https://ci-apps-dev.yunohost.org/ci/job/3624

I normally fixed the linting things which should mean level 7 I hope! 🙏

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3297,7 +3297,6 @@
     },
     "uptime-kuma": {
         "category": "system_tools",
-        "level": 7,
         "subtags": [
             "monitoring"
         ],
--- a/apps.json
+++ b/apps.json
@@ -3301,7 +3301,6 @@
             "monitoring"
         ],
         "state": "working",
-        "maintained": true,
         "url": "https://github.com/YunoHost-Apps/uptime-kuma_ynh"
     },
     "vaultwarden": {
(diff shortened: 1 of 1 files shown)


apps.json
```diff
@@ -3295,6 +3295,18 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/unbound_ynh"
     },
+    "uptime-kuma": {
+        "category": "system_tools",
+        "level": 7,
```
yalh76, line 3300:
```suggestion
```

Level is not needed, it will be calculated automatically

apps.json
```diff
@@ -3295,6 +3295,18 @@
...
+        "subtags": [
+            "monitoring"
+        ],
+        "branch": "master",
+        "revision": "1e274b81a2222208bd5bdf80372ae4d1c989ed98",
```
yalh76, line 3305:
```suggestion
```
As master is the default branch and as it the default head that will be used, thoses parameters are not needed

apps.json
```diff
@@ -3295,6 +3295,18 @@
...
+        "state": "working",
+        "maintained": true,
```
yalh76, line 3307:
```suggestion
```
I's maintained by default

yalh76 (changes requested):
Thanks for adding this package, just need few removal before merging :)

ketsapiwiq:
Done! Thank you!

alexAubin (approved):

PR #1271, opened 2021-11-19
Set phplicensewatcher as working

ericgaspar:
- Set phplicensewatcher as working

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -2390,8 +2390,7 @@
     },
     "phplicensewatcher": {
         "category": "system_tools",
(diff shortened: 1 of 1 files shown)


tituspijean (approved):

tituspijean:
Well I do not see a CI check for that, but let's go. :)

ericgaspar:
> Well I do not see a CI check for that, but let's go. :)

ahah, just like Thomas 😇, we need to see to believe

ericgaspar:
* (Thomas... l'apotre)

PR #1272, opened 2021-11-20
Update app's level

alexAubin:


alexAubin:
Only regression is mailman: https://ci-apps.yunohost.org/ci/job/4024

Not sure to understand why it would fail only during subdir install with that error that seems unrelated

yalh76 (approved):

ericgaspar (approved):

PR #1295, opened 2022-01-02
Armadietto update global level

bencharp:


ericgaspar:
Level are set automatically every friday by the CI.

bencharp:
Ok. Thanks!

PR #1298, opened 2022-01-07
emoncms

ericgaspar:
- Add emoncms to catalog


apps.json
```diff
@@ -761,6 +761,11 @@
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/emailpoubelle_ynh"
     },
+    "emoncms": {
+        "category": "small_utilities",
```
tituspijean, line 765: I'm wondering if `iot` would be a better category.

tituspijean (approved):

PR #1302, opened 2022-01-11
Update homeassistant's level

ewilly:


yalh76:
Not needed, It will be automagically updated next friday

PR #1307, opened 2022-01-14
Retrieve repo url from git remote

tituspijean:
Attempt to solve https://github.com/YunoHost/issues/issues/1963

It retrieves the origin repo url from `.git/config`. Defaults to assumed Github URL if detection fails.

TODO:
- [ ] Handle SSH remotes conversion into https (handle ssh remotes with ports like `ssh://git@gitlab.domainepublic.net:3265/Neutrinet/neutrinet_ynh.git`)
  * https://stackoverflow.com/a/22312124
  * `alias fix-github-origin="git remote set-url --push origin git@github.com:\$(git remote -v | grep \"origin.*push\" | sed 's@https://@@g' | sed 's@\.com@@g' | tr '/:.' ' ' | awk '{print \$3\"/\"\$4}').git"`

### DIFF
--- a/tools/README-generator/make_readme.py
+++ b/tools/README-generator/make_readme.py
@@ -22,7 +22,7 @@ def generate_READMEs(app_path: str):
 
     git = configparser.ConfigParser()
     git.read(app_path / ".git/config")
-    remote = git['remote "origin"']['url']
+    remote = git.get('remote "origin"', {}).get('url', "")
     # TODO: Handle ssh remotes
     remote = re.search("(https:\/\/.*_ynh)\.git", remote)
     if remote is not None:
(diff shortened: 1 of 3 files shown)


tools/README-generator/make_readme.py
```diff
@@ -18,6 +20,14 @@ def generate_READMEs(app_path: str):
     manifest = json.load(open(app_path / "manifest.json"))
     upstream = manifest.get("upstream", {})
 
+    git = configparser.ConfigParser()
+    git.read(app_path / ".git/config")
+    remote = git['remote "origin"']['url']
```
alexAubin, line 25:
```suggestion
    remote = git.get('remote "origin"', {}).get('url', "")
```

tituspijean:
Overcomplicated solution to a problem that will arise when we consider federated forges. Closing for now.

PR #1325, opened 2022-02-09
invoiceninja

ericgaspar:
A `"maintained": false` tag may be less dramatic than just removing it...

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1382,6 +1382,11 @@
         ],
         "url": "https://github.com/YunoHost-Apps/invidious_ynh"
     },
+    "invoiceninja": {
+        "category": "productivity_and_management",
+        "state": "working",
+        "url": "https://github.com/YunoHost-Apps/invoiceninja_ynh"
+    },
     "invoiceninja5": {
         "category": "productivity_and_management",
         "state": "working",


yalh76:
As invoiceninja_ynh has been replaced by invoiceninja5_ynh and as we agreed at the next meeting, keeping invoiceninja_ynh in the catalog is not needed

ericgaspar:
I don't agree.
We should at least ask the maintainer @rndmh3ro what is his plan to migrate from 4 -> 5. 
https://forum.yunohost.org/t/invoiceninja-no-longer-help/18738/7

yalh76:
Maybe there is more informations at https://github.com/YunoHost-Apps/invoiceninja_ynh/issues/12

rndmh3ro:
Copy-Pasting my response from the forums:

>IN4 is still working and can be installed manually. It was pulled from the catalogue and replaced with IN 5 (personally I wouldn’t have pulled it as it still works…). I’ll try to update the app when new versions of IN4 get published.
> That said: IN5 is working now and ready for usage, with many thanks to [@ericg](https://forum.yunohost.org/u/ericg) for helping me with getting it to work. See the announcement here: [Invoice Ninja 5 ready for usage 1](https://forum.yunohost.org/t/invoice-ninja-5-ready-for-usage/18743)


IN4 still gets updates from the developers, the last release was a month ago: https://github.com/invoiceninja/invoiceninja/releases/tag/v4.5.48

In my opinion IN4 should still be installable, updatable and usable in yunohost, as long as the devs still support it. A disclaimer when installing IN4 that says the people should use IN5 is warranted.

yalh76:
wokay, as you want ...

PR #1337, opened 2022-02-22
Add a warning on the README when not on the master branch

Tagadda:
When the default branch from a repo is `testing`, users often see a different "Shipped version" from the version in the catalog, and many ask why.
This PR adds a warning when the README get generated on branch that are not `master`.


tituspijean:
There's an optional `branch` key in `apps.json` to specify if the main branch is not `master`. You should handle that too, as I guess that now `main` has been pushed by Github, some mixups are bound to happen.

tituspijean (approved):
Great idea! 👍

alexAubin (approved):
LGTM, following review with titus, I improved the mechanism to only display the note if the version is different from master, and actually display master's version

PR #1338, opened 2022-02-23
Add Anti-Features in READMEs

Tagadda:


tools/README-generator/make_readme.py
```diff
@@ -53,6 +59,12 @@ def generate_READMEs(app_path: str):
         else:
             disclaimer = None
 
+        # TODO: Add url to the documentation... and actually create that documentation :D
+        antifeatures = [antifeatures_list[a] for a in from_catalog.get('antifeatures', [])]
+        for antifeature in antifeatures:
+            antifeature['title'] = antifeature['title'].get(lang_suffix, 'en')
+            antifeature['description'] = antifeature['description'].get(lang_suffix, 'en')
```
tituspijean, line 66: Wild thought: could we allow to define one's own description in the manifest, if one wants to be more specific than the generic descriptions in `antifeatures.yaml`?

Tagadda: My guess was that more specific details could go under the "Disclaimer" section, and this was a more generic warnings about that "type" of anti-feature, not the app implementation of the anti-feature.
But since this is the app README, that could make sense to have an additional description from the manifest, and hoping that it's not confusing for the one installing the app.

Tagadda:
This is ready, at least! So the anti-features get displayed on the readme.
If the one want a custom description for the antifeature in the README, they could add in the `manifest.json` the following :
```json
{
    "antifeatures": {
        "non-free-addons": {
            "en": "Extensions from the Store might be closed source.",
            "fr": "Les extensions disponibles sur la gallerie peuvent être privatrices."
        }
    }
}
```

tituspijean (approved):

tools/README-generator/make_readme.py
```diff
@@ -53,6 +60,15 @@ def generate_READMEs(app_path: str):
         else:
             disclaimer = None
 
+        # TODO: Add url to the documentation... and actually create that documentation :D
+        antifeatures = { a: antifeatures_list[a] for a in from_catalog.get('antifeatures', [])}
+        for k, v in antifeatures:
+            v['title'] = v['title'].get(lang_suffix, 'en')
+            if manifest.get("antifeatures", {}).get(k, 'en'):
+                v['description'] = manifest.get("antifeatures", {}).get(k, 'en')
```
alexAubin, line 68:
```suggestion
            if manifest.get("antifeatures", {}).get(k):
                v['description'] = manifest.get("antifeatures", {}).get(k).get(lang_suffix, 'en')
```

Maybe a typo here ? Not 100% sure

alexAubin: In fact `.get(lang_suffix, 'en')` is probably not what we want ... we probably want a small helper similar to [this one](https://github.com/YunoHost/yunohost/blob/dev/src/utils/i18n.py#L24) : 

```python
def value_for_lang(values, lang):
     if not isinstance(values, dict):
         return values
     if lang in values:
         return values[lang]
     elif "en" in values:
         return values["en"]
     else:
         return list(values.values())[0]
  ```

File: tools/README-generator/templates/README.md.j2 [hunk 3]
```diff
@@ -49,6 +49,14 @@ If you don't have YunoHost, please consult [the guide](https://yunohost.org/#/in
 {{ disclaimer }}
 {% endif -%}
 
+{% if antifeatures -%}
+## Antifeatures
```
alexAubin, line 53:
```suggestion
## :red_circle: Antifeatures
```

tools/README-generator/templates/README_fr.md.j2
```diff
@@ -31,6 +31,14 @@ Si vous n'avez pas YunoHost, regardez [ici](https://yunohost.org/#/install) pour
 {{ disclaimer }}
 {% endif -%}
 
+{% if antifeatures -%}
+## Fonctions indésirables
```
alexAubin, line 35:
```suggestion
## :red_circle: Fonctions indésirables
```

alexAubin (approved):

PR #1340, opened 2022-02-23
Remove Bozon from catalog

ewilly:
Bozon upstream seems not maintained anymore (last commit on 7 Nov 2018)
What to do with https://github.com/YunoHost-Apps/bozon_ynh ?

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -285,6 +285,18 @@
         ],
         "url": "https://github.com/YunoHost-Apps/borgserver_ynh"
     },
+    "bozon": {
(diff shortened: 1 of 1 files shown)


tituspijean:
We now have the anti-features tags, that will be displayed in the catalog at some point. (this one https://github.com/YunoHost/apps/blob/master/antifeatures.yml#L55-L62 seems fitting)

Let's put this PR as `pending` and close it when the feature is implemented.

ericgaspar (approved):

PR #1371, opened 2022-03-22
Libreddit

ericgaspar:
Add libreddit to catalog


tituspijean (approved):
Approved! The app would be tagged with the "promotes non-free network" label with the future anti-features system. ;)

PR #1410, opened 2022-04-05
Add "django-fritzconnection" to cataglog

jedie:
App package: https://github.com/YunoHost-Apps/django-fritzconnection_ynh
Project sources: https://github.com/jedie/django-fritzconnection

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -632,6 +632,14 @@
         "state": "inprogress",
         "url": "https://github.com/YunoHost-Apps/django-for-runners_ynh"
     },
+    "django-fritzconnection": {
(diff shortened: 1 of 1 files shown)


jedie:
Hm! I created this changes via [add_or_update.py](https://github.com/YunoHost/apps/blob/master/add_or_update.py) ...

tituspijean:
> Hm! I created this changes via [add_or_update.py](https://github.com/YunoHost/apps/blob/master/add_or_update.py) ...

Oh! That's a script I have not seen used for a long time. I think it requires some updates to get up to latest standards. :)

PR #1454, opened 2022-06-01
Remove chtickynotes_ynh

yalh76:
chtickynotes_ynh is a package directly including sources.
Sources are 7 years old:

- https://github.com/YunoHost-Apps/chtickynotes_ynh/tree/master/sources/libs
- https://github.com/YunoHost-Apps/chtickynotes_ynh/tree/master/sources/plugin

initial developper didn't plan to work back on the project: https://github.com/YunoHost-Apps/chtickynotes_ynh/issues/5

The code include google stuff: https://github.com/YunoHost-Apps/chtickynotes_ynh/issues/2

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -349,15 +349,6 @@
         "state": "working",
         "url": "https://github.com/YunoHost-Apps/cheky_ynh"
     },
-    "chtickynotes": {
-        "category": "office",
-        "level": 8,
-        "state": "working",
-        "subtags": [
-            "text"
-        ],
-        "url": "https://github.com/YunoHost-Apps/chtickynotes_ynh"
-    },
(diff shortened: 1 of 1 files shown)


Tagadda (approved):
I think we should create a process to warn admins that some apps are not safe anymore.
Maybe add a new flag `disabled` in the apps.json, or create a `removed.json`.

yalh76:
or something from [antifeatures](https://github.com/YunoHost/apps/pull/1338) ?

alexAubin:
Note that there's already "sort-of" of a warning, because nowadays the diagosis will complain if it finds an app installed that doesn't exist in the catalog (or is in the catalog with level <= 4)

PR #1477, opened 2022-07-15
Update app's level

yalh76:
- [x] [mediawiki](https://ci-apps.yunohost.org/ci/job/8985)
- [ ] [django-fmd](https://ci-apps.yunohost.org/ci/job/8986)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -406,7 +406,7 @@
     },
     "code-server": {
         "category": "dev",
-        "level": 6,
+        "level": 8,
         "state": "working",
         "subtags": [
             "programming"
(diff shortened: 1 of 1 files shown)


yalh76:
@jedie, there is an issue with `django-fmd`

When you added it in apps.json, you added it as `django-find-my-device`, but in the manifest.json, it's `django-fmd`...

Which one we choose as id for the package ?

jedie:
Hm. Sorry. In apps.json it must be the id?

yalh76:
> Hm. Sorry. In apps.json it must be the id?

yes

so, or we change the id to `django-fmd` in apps.json or changing the id in the manifest.json and the name of the repo to `django-find-my-device`

jedie:
> change the id to django-fmd

This sounds the best option. Don't know when I have the time to create a PR

ericgaspar (approved):

PR #1500, opened 2022-08-17
add Minecraft

ericgaspar:
- Adding Minecraft to the catalog (and testing the limits of antifeatures 😅)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -2348,6 +2348,17 @@
         ],
         "url": "https://github.com/YunoHost-Apps/mindmaps_ynh"
     },
+    "minecraft": {
+        "antifeatures": [
+            "not-totally-free"
+        ],
+        "category": "games",
+        "state": "working",
+        "subtags": [
+            "network"
+        ],
+        "url": "https://github.com/YunoHost-Apps/minecraft_ynh"
+    },
     "minetest": {
         "category": "games",
         "level": 8,


jedie (approved):

Tagadda:
Hehe :grin: In my understanding, `not-totally-free` excludes by purpose non-free softwares. The French description states "Application sous licence libre restreinte".
The "not totally" refers to licences that prevent some use case (ie. non-commercial use only...).

ericgaspar:
We could add something like that in `antifeatures.yml`

```
- id: non-free-software
  icon: ???
  title:
    en: "non free software"
    fr: "Logiciel propriétaire"
  description:
    en: "This software contains proprietary software."
    fr: "Ce logiciel contient du code propriétaire."
```

Tagadda:
Feel free to create a PR to discuss it, but I'm not sure that providing non-free softwares in YunoHost main catalog is a good idea.

ericgaspar:
Maybe we should remove minecraft_ynh from the official repo then.

Tagadda:
To clarify, I'm not against YunoHost providing such a package, but I'm against YunoHost providing it *by default*.
One would have to enable a global settings or edit a file as one do on Debian to enable `non-free` repositories.

OniriCorpe:
aren't Spigot and Paper open-source minecraft servers?
maybe we can package them but the official server?

tituspijean:
IIRC they both require the Minecraft server jar file to be patched to generate their own jar files.

And both require the admin to add a setting "proving" they accept Mojang's EULA.

Definitely not free. 🫠

OniriCorpe:
I see 😓

PR #1502, opened 2022-08-18
ffsync is not maintained anymore

tituspijean:


Josue-T:
Well see cf [#38](https://github.com/YunoHost-Apps/ffsync_ynh/pull/38), it's not really the end of ffsync, just didn't have enough time to fix it before bullseye is out.

yalh76 (approved):

tituspijean:
> Well see cf [#38](https://github.com/YunoHost-Apps/ffsync_ynh/pull/38), it's not really the end of ffsync, just didn't have enough time to fix it before bullseye is out.

cf. my answer over there: https://github.com/YunoHost-Apps/ffsync_ynh/pull/40#issuecomment-1220423449

alexAubin:
Merging because we clearly are not going to encourage people using apps with unmaintained upstream and deprecated technologies, and this is only an "antifeature" tag which is not even yet implemented in uis...

PR #1514, opened 2022-09-12
Set Bibliogram as not working

ericgaspar:


yalh76:
Why is it not working ?

ericgaspar:
Instagram feeds are blocked upstream.
More details here:
https://cadence.moe/blog/2022-09-01-discontinuing-bibliogram

PR #1522, opened 2022-10-03
Enh readme current branch

yalh76:
The comment should only be added to the repository default branch when it's not the branch used by apps.json

I've seen the comment added to a testing branch that is not the default branch used by apps.json
You can look at the history of [drupal](https://github.com/YunoHost-Apps/drupal_ynh/commits/master) for example


alexAubin:
Eeeeh I'm confused about the diff because the current code is a feature and not a bug

I guess what you're pointing out for drupal is that the default branch is `master`, so the issue that this is trying to address (= users browsing the README and seeing the "testing" version thinking this is the one released and getting confused with what they see in the catalog) doesnt happen for drupal ...

(And also the fact that this script indeed commits this for any branch that's not master/main, even PRs)

The issue is that, if I remember correctly, a long time ago it was decided that `testing` should be the default branch for apps such that pull requests would get created to `testing` instead of `master`/`main` ... but in practice there's still a lot of apps with `master`/`main` as the default branch ? (I have no quantitative data yet)

So imho we should clarify what we actually want to do. If we go back to "keep `master`/`main` as the default branch and just manually create / move all PRs to `testing`", this solves the issue and we don't need no README tweaking. If we stick to the rule of `testing` as default branch, then we may keep editing the README to prevent user confusion, or find another design.

yalh76:
> The issue is that, if I remember correctly, a long time ago it was decided that `testing` should be the default branch for apps such that pull requests would get created to `testing` instead of `master`/`main` ... but in practice there's still a lot of apps with `master`/`main` as the default branch ? (I have no quantitative data yet)

Actually most of repositories, like more than 90% use master as default branch. 

> 
> So imho we should clarify what we actually want to do. If we go back to "keep `master`/`main` as the default branch and just manually create / move all PRs to `testing`", this solves the issue and we don't need no README tweaking. If we stick to the rule of `testing` as default branch, then we may keep editing the README to prevent user confusion, or find another design.

Fully agree with you, we should discuss at the next meeting of the pro and cons of having master/main or testing branch as the default branch

yalh76:
In addition, with the comments added, we need to resolve conflicts when merging PR from testing to master...
https://github.com/YunoHost-Apps/fab-manager_ynh/pull/29

This PR comes from an autoupdate PR merged to testing.

There is no way, that we need to resolve conflicts for something that worked smoothly before

PR #1555, opened 2022-12-23
Update app's level

ericgaspar:


ericgaspar:
@tituspijean shall we merge this?

PR #1566, opened 2023-01-02
add statping app

oiseauroch:
[Statping-ng](https://github.com/statping-ng) is a lightweight extern monitoring tool.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3396,7 +3396,7 @@
             "Keepeek",
             "Koken",
             "Orkis Ajaris",
-            "Orphéa"
+            "Orph\u00e9a"
         ],
         "state": "working",
         "subtags": [
@@ -3557,7 +3557,6 @@
     "proxitok": {
         "category": "social_media",
         "level": 7,
-        "state": "notworking",
         "potential_alternative_to": [
             "TikTok"
         ],
@@ -3795,7 +3794,7 @@
         "maintained": false,
         "potential_alternative_to": [
             "BitTorrent",
-            "µTorrent®"
+            "\u00b5Torrent\u00ae"
         ],
         "revision": "34ba187c2e76dc6adf77de41304647f6569c6dd8",
         "state": "notworking",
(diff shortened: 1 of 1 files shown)


apps.json
```diff
@@ -3795,7 +3794,7 @@
         "maintained": false,
         "potential_alternative_to": [
             "BitTorrent",
-            "µTorrent®"
+            "\u00b5Torrent\u00ae"
```
alexAubin, line 3797: uuuh bro you fucked up the file encoding :|

apps.json
```diff
@@ -3557,7 +3557,6 @@
     "proxitok": {
         "category": "social_media",
         "level": 7,
-        "state": "notworking",
```
alexAubin, line 3560: also uuuh why change this

apps.json
oiseauroch: I just used the provided script. 
Revert the encoding

apps.json
oiseauroch: Again the script did this. but studying file apps.json, there is both status on the app : 
```
   "proxitok": {
        "category": "social_media",
        "level": 7,
        "state": "notworking",
        "potential_alternative_to": [
            "TikTok"
        ],
        "state": "working",
        "url": "https://github.com/YunoHost-Apps/proxitok_ynh"
    },
```

apps.json
```diff
@@ -4150,6 +4150,12 @@
         "state": "notworking",
         "url": "https://github.com/YunoHost-Apps/staticwebapp_ynh"
     },
+    "statping_ng": {
+        "branch": "master",
+        "revision": "c43ae9a4f32e2cbe3df47210ba9c22558071b11e",
```
alexAubin, line 4155: Cheers !

```suggestion
        "revision": "c43ae9a4f32e2cbe3df47210ba9c22558071b11e",
```

This works but note that we encourage to not set any revision here ... otherwise you'll have to come and update this hash here each time you update the master branch

oiseauroch: Ok. It could be interesting to update the add_or_update.py script because using it give me also revision

apps.json
alexAubin: uuugh indeed x_x

PR #1581, opened 2023-01-06
Update calckey lvl

oufmilo:


tituspijean:
It's automatically updated every Friday night, no need to update it manually. :)

PR #1588, opened 2023-01-09
!testme Add reverseproxy_ynh to catalog

selfhoster1312:
A simple nginx reverse proxy.

The state of this is "works for me" but feedback is welcome to add more features.


selfhoster1312:
Unrelated, but why is redirect_ynh in publishing category? Should it not be in systems_tools/network too?

ericgaspar (approved):

PR #1624, opened 2023-01-29
Add auto-updater upgrader

tituspijean:
Github has deprecated some older versions of `actions/checkout` and `peter-evans/create-pull-request` actions.
Since I'm lazy, let's automate the process.


alexAubin (approved):

tituspijean:
Status: works, but getting some secondary rate limit block sometimes.

PR #1627, opened 2023-02-02
Add initial json schemas for validation of TOML app files (manifest.toml, tests.toml)

Salamandar:
This kinda is a draft.


Salamandar:
This is now complete.

schemas/manifest.schema.json
```diff
@@ -0,0 +1,455 @@
...
+        "maintainers": {
+            "description": "List of maintainers",
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+        },
+        "upstream": {
+            "type": "object",
+            "required": ["license", "website", "code"],
```
alexAubin, line 77:
```suggestion
            "required": ["license"],
```

Unfortunately we can't make those mandatory because some small projects often have no actual website, and (more rarely), we have apps which have no actual code upstream, such as my_webapp, vpnclient, etc.

schemas/manifest.schema.json
```diff
@@ -0,0 +1,455 @@
...
+                        "help": {"$ref": "#/$defs/translated_string"},
+                        "choices": {
+                            "type": "object",
+                            "required": [],
+                            "additionalProperties": false,
+                            "patternProperties": {
+                                "^[a-z][a-z0-9_]*$": {
+                                    "type": "string"
+                                }
+                            }
```
alexAubin, line 163: Unfortunately we should also allow regular list of strings i think ? :grimacing:  Dunno how complex it is to allow both kind of structures

schemas/manifest.schema.json
```diff
@@ -0,0 +1,455 @@
...
+                        "^.[a-z][a-z0-9_]*$": {
+                            "type": "object",
+                            "required": ["url"],
+                            "additionalProperties": false,
+                            "properties": {
+                                "url": {
+                                    "type": "string",
+                                    "pattern": "^/.*$"
+                                },
+                                "show_title": {
```
alexAubin, line 273:
```suggestion
                                "show_tile": {
```

schemas/manifest.schema.json
```diff
@@ -0,0 +1,455 @@
...
+                                "string", "text", "select", "tags", "email", "url", "date", "time",
+                                "color", "password", "path", "boolean", "domain", "user", "group",
+                                "number", "range", "alert", "markdown", "file", "app"
+                            ]
+                        },
+                        "default": {
+                            "type": ["string", "number", "boolean"]
+                        },
+                        "ask": {"$ref": "#/$defs/translated_string"},
+                        "help": {"$ref": "#/$defs/translated_string"},
```
alexAubin, line 154: So eeeh we should also have `example` in here I think, which is often use (though less than before because it's not relevant for many stuff)

alexAubin: Aaaand possibly an `optional` = True or False key ?

schemas/manifest.schema.json
```diff
@@ -0,0 +1,455 @@
...
+                            "type": "object",
+                            "required": [],
+                            "additionalProperties": false,
+                            "properties": {
+                                "default": {
+                                    "type": "integer"
+                                },
+                                "exposed": {
+                                    "type": "boolean"
+                                },
```
alexAubin, line 310: Unfortunately this can be a boolean or `"TCP"` / `"UDP"`

schemas/manifest.schema.json
```diff
@@ -0,0 +1,455 @@
...
+                            }
+                        }
+                    }
+                },
+                "sources": {
+                    "type": "object",
+                    "required": [],
+                    "additionalProperties": false,
+                    "patternProperties": {
+                        "^[a-z][a-z0-9_]*$": {
```
alexAubin, line 323:
```suggestion
                        "^[a-z][a-z0-9_-]*$": {
```

Some sources have `-` in their names

alexAubin (approved):
Feedback after running on all the current manifest.toml in the wild, excited to integrate this in example_ynh and linter o/

schemas/manifest.schema.json
Salamandar: Alright :)

schemas/manifest.schema.json
Salamandar: It's not complex at all, actually I support both string and list of strings for the apt packages already :)

Salamandar: Ah I already fixed it but not pushed yet ^^

schemas/manifest.schema.json
Salamandar: Ooooops :)

schemas/manifest.schema.json
Salamandar: Fixed : 
```
                        "example": {"type": ["string", "number", "boolean"]},
                        "optional": {"type": "boolean"},
```

schemas/manifest.schema.json
Salamandar: *gasp* Alright, I have put a lot of `"^[a-z][a-z0-9_]*$"` in the file. Should i change all of them ? Or only this one ?

schemas/manifest.schema.json
```diff
@@ -0,0 +1,473 @@
...
+                "license": {"type": "string"},
+                "website": {"type": "string", "format": "uri-reference"},
+                "demo": {"type": "string", "format": "uri-reference"},
+                "admindoc": {"type": "string", "format": "uri-reference"},
+                "userdoc": {"type": "string", "format": "uri-reference"},
+                "code": {"type": "string", "format": "uri-reference"},
+                "fund": {"type": "string", "format": "uri-reference"},
+                "cpe": {
+                    "type": "string",
+                    "pattern": "^cpe:[0-9\\.]*:[a-z]*:[a-zA-Z0-9_]*:[a-z]*$"
```
alexAubin, line 89: Hmmm so there's a couple minor issues with this ... I think we don't have the "end part" of the regex which if i recall correctly contains the version number or something ... Also we should allow for `-` in the name

```
In key : upstream > cpe
'cpe:2.3:a:home-assistant:home-assistant' does not match '^cpe:[0-9\\.]*:[a-z]*:[a-zA-Z0-9_]*:[a-z]*$'
In key : upstream > cpe
'???' does not match '^cpe:[0-9\\.]*:[a-z]*:[a-zA-Z0-9_]*:[a-z]*$'
In key : upstream > cpe
'cpe:2.3:a:simplex:simplex_chat' does not match '^cpe:[0-9\\.]*:[a-z]*:[a-zA-Z0-9_]*:[a-z]*$'
In key : upstream > cpe
'cpe:2.3:a:pfsense:pfsense-pkg-wireguard' does not match '^cpe:[0-9\\.]*:[a-z]*:[a-zA-Z0-9_]*:[a-z]*$'
```

I think I would go for a more constraining one for now (anyway this field ain't really used for now) : 

```suggestion
                    "pattern": "^cpe:2.3:a:[a-zA-Z0-9_-]*:[a-zA-Z0-9_-]*$"
```

schemas/manifest.schema.json
```diff
@@ -0,0 +1,473 @@
...
+            }
+        },
+        "integration": {
+            "type": "object",
+            "required": ["yunohost", "architectures", "multi_instance", "ldap", "sso"],
+            "additionalProperties": false,
+            "properties": {
+                "yunohost": {
+                    "type": "string",
+                    "pattern": "^>= ?[0-9]*\\.[0-9]*\\.[0-9]*"
```
alexAubin, line 100:
```suggestion
                    "pattern": "^>= ?[0-9]*\\.[0-9]*\\.[0-9]*"
```

Ideally we should allow for `>= 11.x` or even `>= 11`, or not if too annoying to implement :sweat_smile:

schemas/manifest.schema.json
```diff
@@ -0,0 +1,473 @@
...
+                            "properties": {
+                                "url": {
+                                    "type": "string",
+                                    "pattern": "^/.*$"
+                                },
+                                "show_tile": {
+                                    "type": "boolean"
+                                },
+                                "allowed": {
+                                    "type": "string"
```
alexAubin, line 287: Apparently this can also be a list of allowed groups ... not sure if that's a bug or a feature

schemas/manifest.schema.json
```diff
@@ -0,0 +1,473 @@
...
+                                                "latest_github_tag",
+                                                "latest_github_release",
+                                                "latest_github_commit"
+                                            ]
+                                        },
+                                        "upstream": {
+                                            "type": "string"
+                                        },
+                                        "asset": {
+                                            "oneOf": [
```
alexAubin, line 397: In addition to `strategy`, `upstream` and `asset`, we can have a `force_version` arg of type string, cf for example https://github.com/YunoHost-Apps/indexhibit_ynh/blob/master/manifest.toml#L45

schemas/manifest.schema.json
```diff
@@ -0,0 +1,473 @@
...
+            "type": "string"
+        },
+        "name": {
+            "description": "App Name",
+            "type": "string"
+        },
+        "version": {
+            "description": "App and package version",
+            "type": "string",
+            "pattern": "^[0-9\\.]*~ynh[0-9]*"
```
alexAubin, line 65: Ogod there's a few ones not passing this because they have letters and dashes and stuff ... such as gamja, forgejo 

I was thinking "let's use the one from the linter" but it's absolutely horrendous : https://github.com/YunoHost/package_linter/blob/master/package_linter.py#L379

So idk watdo x_x

schemas/manifest.schema.json
```diff
@@ -0,0 +1,473 @@
...
+                    }
+                },
+                "permissions": {
+                    "type": "object",
+                    "required": [],
+                    "additionalProperties": false,
+                    "patternProperties": {
+                        "^.[a-z][a-z0-9_]*$": {
+                            "type": "object",
+                            "required": ["url"],
```
alexAubin, line 276: Zblerg turns out `url` is not required because sometimes they are only internal / pure-LDAP ... though could also be symptomatic of a typo ...

Not sure what do, but I think I would remove it from the mandatory fields

```suggestion
                            "required": [],
```

alexAubin (approved):
A bunch of other minor comments, I think we can go into production after these even though there'll still be a couple of boring edge cases to handle later

schemas/manifest.schema.json
alexAubin: Hmmmm idk, not hard opinion on this, I'm mainly trying to check what's causing issues in existing TOML :grimacing:

schemas/manifest.schema.json
tituspijean:
```suggestion
                    "pattern": "^>= ?[0-9]*(?:\\.[0-9]*)*"
```
(untested)

Salamandar: I found even better :
`^>= ?[0-9]{1,2}(\.[0-9]{1,2}){0,2}$`

schemas/manifest.schema.json
Salamandar: fixed

schemas/manifest.schema.json
Salamandar: Let's go with a simpler regex: `^[0-9]{1-3}([.-][0-9a-z]*~ynh[0-9]*$`

Salamandar:
All issues should be fixed now.

PR #1655, opened 2023-03-13
POC for new declarative app source auto-update mechanism

alexAubin:
Soooo follow-up of various discussions regarding the maintenance of the app source auto-update mechanism, I drafted this script

Also related to https://github.com/YunoHost/issues/discussions/2164

The point of this script is to be ran by the YunoHost infrastructure every X days and automatically create a pull request on corresponding app (not implemented yet, for now I'm focusing on drafting the grammar and behavior to fetch the appropriate info on the upstream which is the hardest part)

This introduce a new syntax in the `resources.sources` part of the `manifest.toml`

For example, for firefly-iii ("simple" nominal case) this yields something like : 

```toml
        [resources.sources.main]
        url = "https://github.com/firefly-iii/firefly-iii/archive/refs/tags/v6.0.4.tar.gz"
        sha256 = "..."
        autoupdate.strategy = "latest_github_tag"
        autoupdate.asset = "tarball"   # This is the default
```

And for Wireguard (more complex case with a per-architecture asset) ; 

```toml
    [resources.sources]
        [resources.sources.main]
        in_subdir = false
        arm64.url = "https://github.com/ngoduykhanh/wireguard-ui/releases/download/v0.4.0/wireguard-ui-v0.4.0-linux-arm64.tar.gz"
        arm64.sha256 = "..."
        i386.url = "https://github.com/ngoduykhanh/wireguard-ui/releases/download/v0.4.0/wireguard-ui-v0.4.0-linux-386.tar.gz"
        i386.sha256 = "..."
        armhf.url = "https://github.com/ngoduykhanh/wireguard-ui/releases/download/v0.4.0/wireguard-ui-v0.4.0-linux-arm.tar.gz"
        armhf.sha256 = "..."
        amd64.url = "https://github.com/ngoduykhanh/wireguard-ui/releases/download/v0.4.0/wireguard-ui-v0.4.0-linux-amd64.tar.gz"
        amd64.sha256 = "..."

        autoupdate.strategy = "latest_github_release"
        autoupdate.asset.amd64 = ".*linux-amd64.tar.gz"
        autoupdate.asset.i386 = ".*linux-386.tar.gz"
        autoupdate.asset.armhf = ".*linux-arm.tar.gz"
        autoupdate.asset.arm64 = ".*linux-arm64.tar.gz"
```

So basically the `autoupdate.strategy` defines how to look for the newest version. Two values are supported for now (and coupled to Github which is like 90% of apps :grimacing: ) 
- `latest_github_tags` : look for the latest tag (by sorting tags and finding the "largest" version). Then using the corresponding tar.gz url. Tags containing `rc`, `beta`, `alpha`, `start` are ignored, and actually any tag which doesn't look like `x.y.z` or `vx.y.z`
- `latest_github_release` : similar to `latest_github_tags`, but starting from the list of releases. Pre- or draft releases are ignored. Releases may have assets attached to them, in which case you can define: 
    - `autoupdate.asset = "some regex"` (when there's only one asset to use). The regex is used to find the appropriate asset among the list of all assets
    - or several `autoupdate.asset.$arch = "some_regex"` (when the asset is arch-specific). The regex is used to find the appropriate asset for the specific arch among the list of assets (cf example for wireguard)

So "of course" this doesn't cover all cases ... *maybe* we could have a similar mechanism to check for tags on any git repo in a forge-agnostic way, maybe using `git ls-remote`. But the tarball URL is gonna be forge-specific. Maybe Gitlab is easy to support.

Some cases (maybe 5~10ish percent of apps?) are clearly not going to be generalizable, for those I guess it's okay-ish to keep a github workflow and see later what we can do

#### TODO once we validate the proposed syntax

- [x] Actually propagate the info on the `manifest.toml`. I'm probably gonna go for some brutal `sed` instead of trying to parse/dump the toml "properly" because it doesn't seem straightforward at all to not mess comments and indentation ... 
- [x] If there's a diff, create the corresponding pull request
- [ ] Create the corresponding cron job on our infra ... Choose how often we run it etc, maybe we should end up defining an `autoupdate.frequency` parameter in the manifest.toml if once per day is too much for some apps

### DIFF
--- a/tools/autoupdate_app_sources/autoupdate_app_sources.py
+++ b/tools/autoupdate_app_sources/autoupdate_app_sources.py
@@ -1,0 +1,125 @@
+import re
+import sys
+import requests
+import toml
+import os
+
+STRATEGIES = ["latest_github_release", "latest_github_tag"]
+
+GITHUB_LOGIN = open(os.path.dirname(__file__) + "/../../.github_login").read().strip()
+GITHUB_TOKEN = open(os.path.dirname(__file__) + "/../../.github_token").read().strip()
+
+
+def filter_and_get_latest_tag(tags):
+    filter_keywords = ["start", "rc", "beta", "alpha"]
+    tags = [t for t in tags if not any(keyword in t for keyword in filter_keywords)]
+
+    for t in tags:
+        if not re.match(r"^v?[\d\.]*\d$", t):
+            print(f"Ignoring tag {t}, doesn't look like a version number")
+    tags = [t for t in tags if re.match(r"^v?[\d\.]*\d$", t)]
+
+    tag_dict = {t: tag_to_int_tuple(t) for t in tags}
+    tags = sorted(tags, key=tag_dict.get)
+    return tags[-1]
+
+
+def tag_to_int_tuple(tag):
+
+    tag = tag.strip("v")
+    int_tuple = tag.split(".")
+    assert all(i.isdigit() for i in int_tuple), f"Cant convert {tag} to int tuple :/"
+    return tuple(int(i) for i in int_tuple)
+
+
+class AppAutoUpdater():
+
+    def __init__(self, app_path):
+
+        if not os.path.exists(app_path + "/manifest.toml"):
+            raise Exception("manifest.toml doesnt exists?")
+
+        manifest = toml.load(open(app_path + "/manifest.toml"))
+
+        self.current_version = manifest["version"].split("~")[0]
+        self.sources = manifest.get("resources", {}).get("sources")
+
+        if not self.sources:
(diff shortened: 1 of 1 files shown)


mh4ckt3mh4ckt1c4s:
I don't know if it is the only one in this case but the [SearXNG app](https://github.com/YunoHost-Apps/searxng_ynh) is rolling-release (i.e., each commit on master is a release), and there are no tags or github release. So the methods `latest_github_tags` or `latest_github_release` won't work. If it is SearXNG-specific, I'll deal with it in its repo directly, but if it is more general, another option would be welcome.

alexAubin:
Yes, we ain't gonna achieve 100% use case coverage .. But we could implement something like `latest_github_commit`

PR #1676, opened 2023-05-22
add nonbiri to catalog

ericgaspar:


tituspijean (approved):
I would make the Overview section of the README impersonal, though:

> Just something I made for myself because MangaDex's new UI is bloated piece of hot garbage.

This is a tad too personal and judgy, we are only packagers here. :)

PR #1685, opened 2023-06-10
Add requires-its-own-domain antifeature.

Salamandar:
This is for apps that only support installation on a domain root, like cryptpad, fider, vaultwarden, etc.

### DIFF
--- a/antifeatures.toml
+++ b/antifeatures.toml
@@ -88,3 +88,11 @@ title.en = "Not totally free upstream"
 title.fr = "Application sous licence libre restreinte"
 description.en = "The packaged app is under an overall free licence, but with clauses that restrict its use."
 description.fr = "L'application packagée est sous une licence globalement libre, mais avec des clauses qui pourraient restreindre son utilisation."
+
+[requires-its-own-domain]
+icon = "globe"
+title.en = "Requires its own (sub) domain"
+title.fr = "Nécessite son propre (sous) domaine"
+description.en = "This app uses its own (sub) domain that you first need to create in the admin panel."
+description.fr = "Cette application nécessite son propre (sous) domaine que vous devez créer au préalable."
+


tituspijean:
Mmmmh why not! :D

tituspijean (approved):

Tagadda:
I'm not sure that an anti-feature is the best fit for this situation. This information is already given by `full_domain` in `manifest.toml` (example https://github.com/YunoHost-Apps/owncast_ynh/blob/ddc7df21717de97798c6870b7bfd346095ad4565/manifest.toml#L34).

I think a better way to address this issue could be allowing to add a new domain during an app install.

tituspijean:
> I think a better way to address this issue could be allowing to add a new domain during an app install.

Oh yesssss

alexAubin:
> This information is already given by full_domain in manifest.toml (example https://github.com/YunoHost-Apps/owncast_ynh/blob/ddc7df21717de97798c6870b7bfd346095ad4565/manifest.toml#L34).

Eeeeeh I think that info is actually not being used :sweat_smile: Despite that this was discussed several time to add such a mechanism ... I'm not sure exactly

Nevertheless there is a piece of code which "magically" guess that the app is a full-domain one, typically because there's only one domain question (+ the permission url in packaging-v2 is a string, not None) : https://github.com/YunoHost/yunohost/blob/dev/src/app.py#L2785

But maybe the current UX is not great in the webadmin ?

Salamandar:
> This information is already given by `full_domain` in `manifest.toml`

Oh well that's missing in the doc :'(

Salamandar:
> But maybe the current UX is not great in the webadmin ? Like it's only checked after clicking the "install" button ?

Yeah that's exactly why I did this pull request : it's to have the information *before* starting the install.

Tagadda:
> > But maybe the current UX is not great in the webadmin ? Like it's only checked after clicking the "install" button ?
> 
> Yeah that's exactly why I did this pull request : it's to have the information _before_ starting the install.

I agree that the information should be displayed before we install the app somehow, but I don't know if it's really convenient. Once on the catalog, you'll see that some apps require a full domain, so: you exit the catalog, add a domain and then go back to the catalog.

I think allowing to add a domain from the app install step is more straight forward and better UX.
In the webadmin there could be a "New domain" choice in the select input that would create a text input below or something.
In CLI we would be able to type the domain directly and be prompted for confirmation that we do want to add a new domain.

Salamandar:
Yes, agreed.

Salamandar:
Closing because of reasons

PR #1692, opened 2023-07-06
add appflowy

babolab:
Good evening,

My level does not allow me to create a package for ynh, but I think this app is really interresting, for it provides an alternative to notion, which is very powerful.
Please forgive this pull request, not very regular.
Merci d'avance,
Geo


Tagadda:
AppFlowy is a desktop software and does not have a web interface at the moment.
See https://github.com/AppFlowy-IO/AppFlowy/issues/89

PR #1707, opened 2023-08-04
Update apps.toml - Added streams

dragondaddy:


dragondaddy:
url has to be changed after transfer to YunoHost-Apps

PR #1716, opened 2023-08-12
Add InfluxDB v2 to Catalog

M3t4Data:
Tests are OK in my local dev environment


tituspijean:
How would this interact with other apps relying on InfluxDB, like Grafana, NetData, Scrutiny?

tituspijean (approved):

PR #1728, opened 2023-09-08
Add Froxlor to catalog

ericgaspar:


tituspijean (commented):
I am a bit dubious about how this won't clash with our own features and how we handle them.

PR #1742, opened 2023-09-24
Fix manifest schema

Salamandar:
Those are the last fixes to properly validate manifest.toml files and remove all the false negatives.


alexAubin:
Cheers, didnt review all the stuff carefully but looks legit and blindly trusting :+1:

PR #1743, opened 2023-09-24
schemas: Add schemas for apps.toml and categories.toml

Salamandar:
Eh, I don't really know if they could be useful. Maybe in github merge requests checks ?
But well, they were easy to write.

### DIFF
--- a/schemas/apps.toml.schema.json
+++ b/schemas/apps.toml.schema.json
@@ -1,0 +1,64 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://github.com/YunoHost/apps/blob/master/schemas/apps.toml.schema.json",
+    "title": "Yunohost's apps.toml schema",
+    "version": "0",
+
+    "type": "object",
+    "required": [],
+    "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


alexAubin:
Hmmmyes indeed it would be pretty cool to have a github action with this check (there's already one but only checks that it's a valid toml and that there's consistency in the url : https://github.com/YunoHost/apps/blob/master/.github/workflows/main.yml#L21 )

Salamandar:
I think that json-schema can even incorporate some custom logic that can replace the catalog_linter.py (but is it what we want ?)

alexAubin:
I'm re-reading the catalog linter and it's only for : 
- checking the url of the app ends with `{app_id}_ynh` for consistency
- that the `category` info exists and is effectively defined in `categories.yml`

not sure if all of this can be done using the schema but it would be nice, maybe checking that the category exists in `categories.yml` is not really mandatory

Salamandar:
Actually this was more advance in my memory than it is really, so, we can check it's `.*_ynh` but not `{app}_ynh`, so the catalog linter is still useful… by the way i'm rewriting it :D

PR #1746, opened 2023-09-24
schemas: fix tests.v1, args can be booleans too

Salamandar:


Salamandar:
I also pushed a commit that fixes catalog_linter… sorry ><'

PR #1754, opened 2023-09-26
Mark Castopod as broken

xplosionmind:


ericgaspar:
there is `working` / `notworking` but no `broken`

xplosionmind:
Sorry @ericgaspar, I got the syntax wrong.

Replying to @alexAubin, in order to see why the app does not work just look at [the issues](https://github.com/YunoHost-Apps/castopod_ynh/issues)… almost all of them note that the upgrade fails or that there is some PHP error…

In particular: https://github.com/YunoHost-Apps/castopod_ynh/issues/81

Tagadda:
I did try to install the app right now and it's seems to be working good.
You have to go through the post-install process by visiting the app (without /cp-admin)

xplosionmind:
> I did try to install the app right now and it's seems to be working good.
> You have to go through the post-install process by visiting the app (without /cp-admin)

Mmmh it does not really work for me… I think my only option is reinstalling the app. Thanks!

> I'll close this PR for now. If you encounter any problem while installing or using the app, please open a new issue on the app repository!

Yes, I will. Thanks again.

PR #1756, opened 2023-09-26
External links in catalog to open in a new tab

tituspijean:
Hopefully I'm not breaking anything...


Tagadda (approved):
Looks good! I think we should do this on the wishlist too

PR #1787, opened 2023-10-06
Emojiwall for Owncast

the7thNightmare:
Emojiwall for Owncast.\
For the app id though, can I just use `emojiwall` or should I be more specific?


Tagadda:
If it can't be used without Owncast, I would suggest `owncast-emojiwall`
Also, please consider to transfert your repo to YunoHost-Apps organization! This way you'll be able to run the CI against your repo :)

ericgaspar (approved):

PR #1848, opened 2023-11-04
source autoupdater: Use semver for version parsing.

orhtej2:
We currently manually parse versions in format `x.y.z`. Semver does this for us in (hopefully) less error prone way.

One caveat is that [rustdesk](https://github.com/rustdesk/rustdesk-server/releases) and probably others don't strictly follow semver, making `1.0.0a` an invalid version and `1.0.0-1` less than `1.0.0` (`1.0.0-1` is considered pre-release).

### DIFF
--- a/tools/autoupdate_app_sources/autoupdate_app_sources.py
+++ b/tools/autoupdate_app_sources/autoupdate_app_sources.py
@@ -6,6 +6,7 @@ import requests
 import toml
 import os
 import glob
+from semver.version import Version
 from datetime import datetime
 
 STRATEGIES = ["latest_github_release", "latest_github_tag", "latest_github_commit"]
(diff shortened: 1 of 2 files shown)


alexAubin:
So hmmm that sounds interesting but testing it on the battefield I get many issues ... for example many apps have tags starting with `v`. That one is easy-ish and can be solved with a `.strip('v')`

But then I find: 

```
Ignoring tag 1.2.5.1, doesn't look like a version number
Ignoring tag 1.2, doesn't look like a version number
```

(apparently we must have a strict `x.y.z`, `x.y` or `x.y.z.a` being invalid)

and even: 

```
Ignoring tag 23.08.3, doesn't look like a version number
Ignoring tag 2023.04.04, doesn't look like a version number
```

(i.e. semver not happy with zero-padded numbers)

orhtej2:
Thanks for testing, indeed seems semver is even stricter than I thought when it comes to what's a real version.

PR #1920, opened 2023-12-18
Graveyard and drafts

alexAubin:
- Move long-term not-working apps to a new graveyard.toml
- Add a `draft = <repo url>` in the wishlist.toml for apps with an existing (even though notworking repo)


tituspijean:
Let's not forget to document what the graveyard entails for the end-user. 😅

Tagadda (approved):

alexAubin:
>Let's not forget to document what the graveyard entails for the end-user. 😅

Yeah i don't know what to do besides adding a description in the README (cf previous commit), it doesnt really impact users because notworking apps where already not displayed to users anyway, but putting them in graveyard.toml is more like "the app is very unlikely to be revived" ... Though we could also question why not just remove them from apps.toml instead of adding this new file x_x

PR #1927, opened 2023-12-22
Add iceshrimp to catalog

oufmilo:


apps.toml
```diff
@@ -1527,6 +1527,11 @@ state = "working"
 subtags = [ "programming" ]
 url = "https://github.com/YunoHost-Apps/icecoder_ynh"
 
+[iceshrimp]
+category = "social_media"
+potential_alternative_to = [ "X", "Threads", "Mastodon", "Pleroma", "Misskey", "Calckey" ]
+url = "https://github.com/YunoHost-Apps/iceshrimp_ynh"
+
```
ericgaspar, line 1534: is it working? 
```
if so state = "working"
else
state = "notworking"
```

ericgaspar (changes requested):

oufmilo:
not working yet

apps.toml
oufmilo: Not working for the moment

ericgaspar (approved):

PR #1937, opened 2023-12-27
cjdns working status

dkoukoul:
Hello and happy holidays everyone.
Last test show as failed https://ci-apps-dev.yunohost.org/ci/job/12055
but it passed to level 4. It is now clear for me if it is level 6 since it has not linter errors just warnings and the app is a community git repo but I guess you know better what level you should put it as.
Thanks for the help

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -388,7 +388,7 @@ url = "https://github.com/YunoHost-Apps/civicrm_drupal7_ynh"
 
 [cjdns]
 category = "system_tools"
-state = "notworking"
+state = "working"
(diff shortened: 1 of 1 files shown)


ericgaspar (approved):

dkoukoul:
Hello @ericgaspar , thanks for your help with this.
I just noticed after local testing that the cjdns service now can not create the tun0 device anymore. I guess this is due to the restrictions in the systemd.service. So although it is running it is not working as expected. So trying now to figure out which off these capabilities are actually needed so as not to remove everything...

PR #1966, opened 2024-01-17
Added MotionEye

SolAZDev:
Adds [MotionEye](https://github.com/YunoHost-Apps/motioneye_ynh) to the app catalog.
It's still inprogress, but the CI tests passed and my hope is that things will be ready for the next release.

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -2060,6 +2060,11 @@ level = 8
 state = "working"
 url = "https://github.com/YunoHost-Apps/mosquitto_ynh"
 
+[motioneye]
+category = "iot"
+state = "inprogress"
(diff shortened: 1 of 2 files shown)


apps.toml
```diff
@@ -2060,6 +2060,11 @@ level = 8
 state = "working"
 url = "https://github.com/YunoHost-Apps/mosquitto_ynh"
 
+[motioneye]
+category = "iot"
+state = "inprogress"
```
alexAubin, line 2065:
```suggestion
state = "working"
```

considering the only issue left seems to be the linter reporting the app as not in the catalog, let's set it directly as working ?

alexAubin (approved):

PR #1979, opened 2024-01-26
Use direct API request for projects instead of search

Salamandar:


alexAubin (approved):
not really sure what it's about but LGTM :grimacing:

PR #1993, opened 2024-02-05
Set coin as working

Salamandar:


Salamandar:
See https://ci-apps-dev.yunohost.org/ci/job/13292

PR #1999, opened 2024-02-07
Rework list_builder.py

Salamandar:
FAAAAAASTEEEEEEER

i'm not kidding:
Before: ~3m30
After: ~13s


Salamandar:
FYI the slowest part of the code is the "git log for first commit that introduced this app".

tools/list_builder.py
```diff
@@ -3,27 +3,34 @@
...
 
+from app_caches import apps_cache_update_all, app_cache_folder  # pylint: disable=import-error
 from packaging_v2.convert_v1_manifest_to_v2_for_catalog import \
     convert_v1_manifest_to_v2_for_catalog  # pylint: disable=import-error
 
+from appslib.utils import (REPO_APPS_ROOT,  # pylint: disable=import-error
+                           get_catalog, git_repo_age)
+
+# Automatically enables error-to-xmpp
+import appslib.xmpplogger  # pylint: disable=import-error
```
alexAubin, line 26: Kind of nitpicking but ideally we should pick a protocol-agnostic name for this like just "notifications" or idk ... The script is still called `sendxmpppy` but it's historical, under the hood it's actually calling matrix-commander :grimacing:

Salamandar: okay :)

Salamandar: Done: logging_sender

PR #2038, opened 2024-02-18
Wishlist adding: Many texts improvements, to enhance user understanding

OniriCorpe:
I've changed several sentences to make them more eloquent and added links

Idk if it’s easy, but it would be nice to add the catalog on the weblate, so it would be translated in another languages…


OniriCorpe:
Btw i can’t translate for now, as i can’t compile the translation file

store/README.md
```diff
@@ -18,6 +18,7 @@ mkdir -p ../builds/default/v3/
 curl https://app.yunohost.org/default/v3/apps.json > ../builds/default/v3/apps.json
 
 # You will also want to run list_builder.py to initialize the .apps_cache (at least for a few apps, you can Ctrl+C after a while)
+pip3 install tqdm GitPython
```
Psycojoker, line 21: Wouldn't it make more sens in a requirements.txt? Also with a major version pinned to be future proof.

OniriCorpe: Note that the `list_builder.py` is not part of the store app

Tagadda (approved):

PR #2044, opened 2024-02-20
Add penpot to catalog

orhtej2:
Waiting for CI.....


apps.toml
```diff
@@ -2465,6 +2465,11 @@ state = "working"
 subtags = [ "website" ]
 url = "https://github.com/YunoHost-Apps/pelican_ynh"
 
+[penpot]
+category = "productivity_and_management"
```
OniriCorpe, line 2469: i think `dev` would be more suitable

logos/penpot-logo-icon.png
ericgaspar, line 1: This should be named `penpot.png`(there is already an existing logo file in `logos` folder )

apps.toml
ericgaspar: `productivity_and_management` seems fine to me.

logos/penpot-logo-icon.png
orhtej2: 🤦

apps.toml
OniriCorpe: But it’s literally a dev tool, to design apps or websites x)

Salamandar: Well i kinda disagree, i'd say it's closer to a drawing too than to a development tool ^^''

ericgaspar (approved):

PR #2045, opened 2024-02-21
store: init new locales for weblate

OniriCorpe:
https://github.com/YunoHost/issues/issues/2335


kay0u:
I'm not really sure we need to create "empty files", weblate should be able to do it when it's needed

kay0u (approved):

PR #2096, opened 2024-03-06
Fix regex to replace version in manifest

Salamandar:
This regex is used to find (textually, not via toml parser) where to replace the version number in the Manifest.

Until now it was only supporting version numbers containing dots and digits. I changed that to version numbers containing `anything but ~ and "` to handle version numbers with dashes, or manually written version numbers that might be invalid (if someone litteraly writes `version = "x.y~ynh1"). We don’t actually care about the version number we match against because we just replace it.


Psycojoker:
Is it possible than you give an example to justify this change? It's hard to review out of context 😅

Psycojoker (approved):

PR #2098, opened 2024-03-07
ci: add autoblacks actions

Psycojoker:
This add the autoblack action from moulinette on this repository (while adapting the branch name).

This action automatically generate a pull request to format the code with black so we don't lose time by having to run it manually when needed.


OniriCorpe (approved):
lets go

PR #2104, opened 2024-03-07
add the schema link in toml files

OniriCorpe:
notes:
- `graveyard.toml` does not have a schema -> fixed
- `antifeatures.toml` does not have a schema -> fixed
- `Additional properties are not allowed ('draft' was unexpected)` errors for `wishlist.toml` -> fixed

PR status: ready


Psycojoker:
Are those schemas used by linter tools? Do we have a CI job that used them before merging or do we need one?

OniriCorpe:
> Are those schemas used by linter tools? Do we have a CI job that used them before merging or do we need one?

yes it's used by an automatic linter, but only for `apps.toml`
https://github.com/YunoHost/apps/blob/68741582d0539f53fd40ad675f72773231c4a7e8/tools/catalog_linter.py#L14

it's also used by my linter in my IDE, so i can easilly spot errors like:
> `Additional properties are not allowed ('draft' was unexpected)` errors for `wishlist.toml` 

and fix them by hand

even if they're not used by an automatic CI, it's a good enhancement i think:
- it brings standardization (all our .toml files have their associated schema)
- it's good practice for toml
- we can quickly see manually what is broken, it's not perfect but better than nothing

Psycojoker:
> even if they're not used by an automatic CI, it's a good enhancement i think

I think too, I'm interested about how to automatized it and adding it where it's useful like in some CI jobs or equivalent

Psycojoker (approved):

PR #2110, opened 2024-03-09
Flag a lot of apps as deprecated. Add an alternative in the wishlist.

Salamandar:
I manually went to every upstream repository to check that is was indeed long-time deprecated softwares.

Feel free to tell me if you think X or Y should not be flagged as such.

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -37,6 +37,7 @@ level = 6
 state = "working"
 subtags = [ "microblogging", "pictures" ]
 url = "https://github.com/YunoHost-Apps/acropolis_ynh"
+antifeatures = ["deprecated-software"]
 
 [actual]
 category = "productivity_and_management"
@@ -59,6 +60,7 @@ level = 8
 state = "working"
 subtags = [ "db" ]
 url = "https://github.com/YunoHost-Apps/adminer_ynh"
+antifeatures = ["deprecated-software"]
 
 [aeneria]
 category = "iot"
@@ -73,6 +75,7 @@ potential_alternative_to = [ "Google Agenda", "Microsoft Outlook" ]
 state = "working"
 subtags = [ "calendar" ]
 url = "https://github.com/YunoHost-Apps/agendav_ynh"
+antifeatures = ["deprecated-software"]
 
 [agora]
 category = "wat"
@@ -185,6 +188,7 @@ category = "iot"
 level = 8
 state = "working"
 url = "https://github.com/YunoHost-Apps/beehive_ynh"
+antifeatures = ["deprecated-software"]
 
 [biboumi]
 category = "communication"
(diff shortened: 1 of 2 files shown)


CodeShakingSheep:
As discussed here https://github.com/YunoHost/apps/pull/1710 YNH invoiceninja package is still deprecated as it refers to version 4 which isn't developed anymore. Version 5  which is still actively developed is packaged here https://github.com/YunoHost-Apps/invoiceninja5_ynh

Salamandar:
> As discussed here #1710 YNH invoiceninja package is still deprecated as it refers to version 4 which isn't developed anymore. Version 5 which is still actively developed is packaged here https://github.com/YunoHost-Apps/invoiceninja5_ynh

Argh okay, I saw the version 5 in the upstream repo mentioned by invoiceninja_ynh so i didn’t see there was an invoiceninja5_ynh app :)

MadMan247:
This looks good and accurate, why isn't it merged? The only task I could further think of is going through the package list and determining unmaintained, but that would be a separate marathon... Maybe I'll do that.

apps.toml
```diff
@@ -199,6 +203,7 @@ level = 7
 state = "working"
 subtags = [ "books" ]
 url = "https://github.com/YunoHost-Apps/bicbucstriim_ynh"
+antifeatures = ["deprecated-software"]
```
alexAubin, line 206: new repo seems to be maintained ? cf https://github.com/YunoHost-Apps/BicBucStriim_ynh/pull/17

apps.toml
```diff
@@ -1354,6 +1369,7 @@ category = "system_tools"
 level = 7
 state = "working"
 url = "https://github.com/Yunohost-Apps/httpsh_ynh"
+antifeatures = ["deprecated-software"]
```
alexAubin, line 1372: (one commit last month but otherwise inactive, idk if we put the flag or not, it's still more modern than shellinabox?)

alexAubin (approved):
LGTM

apps.toml
alexAubin:
```suggestion
url = "https://github.com/YunoHost-Apps/bicbucstriim_ynh"
```

Salamandar: Yes indeed, might have been an automatic flag in this pr.

apps.toml
Salamandar: Wooopsies, shouldn't have flagged this one ! It's not deprecated at all :
* the core "smart" code is actually `ttyd` and it's maintained
* httpsh is basically a 6 line bash script that calls `login` and ssh.

I could even copy this bash script inside the app package and change the manifest upstream url to ttyd… dunno

alexAubin:
```suggestion
url = "https://github.com/Yunohost-Apps/httpsh_ynh"
```

PR #2120, opened 2024-03-13
readme-generator: only pick files (no folder) on the root of 'screenshots'

OniriCorpe:
## The problem

if a subfolder exists, it is used as a screenshot file:
`Screenshot of Some App`

this is obviously wrong

## The solution

- adapt `readme-generator` to only pick actual files (no folder) on the root of the 'screenshots' directory
- also moved the `screenshots` code part to only run it once, since it do not change with language

this also permit to create subfolders to store pictures only used in docs

## PR status

ready to review

### DIFF
--- a/tools/readme_generator/make_readme.py
+++ b/tools/readme_generator/make_readme.py
@@ -87,9 +87,9 @@ def generate_READMEs(app_path: Path):
             # only pick files (no folder) on the root of 'screenshots'
             for entry in os.scandir(os.path.join(app_path, "doc", "screenshots")):
                 if os.DirEntry.is_file(entry):
-                    screenshots.append(os.path.relpath(entry.path, app_path))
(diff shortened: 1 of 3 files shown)


tools/readme_generator/make_readme.py
```diff
@@ -51,6 +51,16 @@ def generate_READMEs(app_path: Path):
 
     env = Environment(loader=FileSystemLoader(Path(__file__).parent / "templates"))
 
+    screenshots: List[str]
+    screenshots = []
```
Salamandar, line 55: Why do this on two separate lines ?

Salamandar (approved):

tools/readme_generator/make_readme.py
OniriCorpe: I did this by default as I didn't know if python 3.9 supported it

PR #2129, opened 2024-03-14
Fix readme_generator/webhook: make_readme now uses pathlib

Salamandar:


tituspijean:
yolotested in prod, an approved.

PR #2135, opened 2024-03-15
logging_sender.py: permit to choose a matrix chan where to send the notification

OniriCorpe:
## The problem

- Autoupdater pushes it's notifications on yunohost-dev, where it's irrelevant, this notif is for packagers

## The solution

- Permit to choose a matrix channel where to send the notification, so the tool is more versatile and the notifications are pushed to a chan where they're relevant

took inspiration here: https://github.com/YunoHost/webhooks/blob/master/server.py#L34

## PR status

not tested
ready to review

### DIFF
--- a/tools/appslib/logging_sender.py
+++ b/tools/appslib/logging_sender.py
@@ -1,16 +1,39 @@
 #!/usr/bin/env python3
 
 import subprocess
-from shutil import which
 import logging
 import logging.handlers
 
 
-def send_to_matrix(message: str) -> None:
-    if which("sendxmpppy") is None:
-        logging.warning("Could not send error via xmpp.")
-        return
-    subprocess.call(["sendxmpppy", message], stdout=subprocess.DEVNULL)
+def notify(message, channel):
+    print(f"{channel} -> {message}")
+
+    chan_list = ["dev", "apps", "doc"]
+
+    try:
+        any(channel in x for x in chan_list)
+    except False:
+        logging.warning(
+            f"Provided chan '{channel}' is not part of the available options ('dev', 'apps', 'doc')."
+        )
+
+    for char in ["'", "`", "!", ";", "$"]:
+        message = message.replace(char, "")
+
+    try:
+        subprocess.call(
+            [
+                "/var/www/webhooks/matrix-commander",
+                "--markdown -m '{message}' -c /var/www/webhooks/credentials.json --store /var/www/webhooks/store --room 'yunohost-{channel}'",
+            ],
+            stdout=subprocess.DEVNULL,
+        )
+    except subprocess.CalledProcessError as e:
+        logging.warning(
+            f"""Could not send a notification on {channel}.
+            Message: {message}
+            Error: {e}"""
+        )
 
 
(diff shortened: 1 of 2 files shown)


tools/appslib/logging_sender.py
```diff
@@ -1,16 +1,37 @@
...
 import logging
 import logging.handlers
 
 
-def send_to_matrix(message: str) -> None:
-    if which("sendxmpppy") is None:
-        logging.warning("Could not send error via xmpp.")
-        return
-    subprocess.call(["sendxmpppy", message], stdout=subprocess.DEVNULL)
+def notify(message, channel):
```
Salamandar, line 8: I would prefer keeping the more verbose function name send_to_matrix

tools/appslib/logging_sender.py
```diff
@@ -1,16 +1,37 @@
...
+
+    chan_list = ["dev", "apps", "doc"]
+
+    if not any(channel in x for x in chan_list):
+        logging.error(
+            f"Provided chan '{channel}' is not part of the available options ('dev', 'apps', 'doc')."
+        )
+
+    for char in ["'", "`", "!", ";", "$"]:
+        message = message.replace(char, "")
```
Salamandar, line 19: Hmmmm why is that ? Couldn't we escape them ? Or store the message in a file passed to matrix-commander to prevent this kind of issue ?

tools/appslib/logging_sender.py
```diff
@@ -1,16 +1,37 @@
...
+
+    try:
+        subprocess.call(
+            [
+                "/var/www/webhooks/matrix-commander",
+                "--markdown -m '{message}' -c /var/www/webhooks/credentials.json --store /var/www/webhooks/store --room 'yunohost-{channel}'",
+            ],
+            stdout=subprocess.DEVNULL,
+        )
```
Salamandar, line 28:
```suggestion
        subprocess.call([
                "/var/www/webhooks/matrix-commander",
                "--markdown", 
                "-m", message, 
                "-c", "/var/www/webhooks/credentials.json", 
                "--store", "/var/www/webhooks/store",
                "--room", f"yunohost-{channel}",
            ],
            stdout=subprocess.DEVNULL,
        )
```

tools/appslib/logging_sender.py
OniriCorpe: don't know, took from there https://github.com/YunoHost/webhooks/blob/master/server.py?rgh-link-date=2024-03-15T03%3A43%3A50Z#L34

tools/appslib/logging_sender.py
OniriCorpe: not me, I prefer not to name a protocol to be able to notify on several if needed later

Salamandar: Alright, fine with me.

tools/appslib/logging_sender.py
Salamandar: weird, but ok…

Salamandar (approved):

PR #2152, opened 2024-03-21
permit regexes for permissions.url in manifest v2 schema

OniriCorpe:
to avoid this:

>  ⓘ manifest
> 
>     - Error validating manifest using schema: in key resources > permissions > api > url
>        're:__DOMAIN__/control' does not match '^(__[A-Z_]*__)?/.*$'

the regex is really dumb when it detects `re:` (`(re:.*)?`), maybe we can do better?

### DIFF
--- a/schemas/manifest.v2.schema.json
+++ b/schemas/manifest.v2.schema.json
@@ -3,14 +3,15 @@
     "$id": "https://raw.githubusercontent.com/YunoHost/apps/master/schemas/manifest.v2.schema.json",
     "title": "Yunohost app manifest.toml schema",
     "version": "0",
-
     "type": "object",
     "$defs": {
         "translated_string": {
             "anyOf": [
                 {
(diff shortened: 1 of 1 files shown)


schemas/manifest.v2.schema.json
```diff
@@ -300,7 +413,10 @@
                             "required": [],
                             "additionalProperties": false,
                             "properties": {
-                                "url": {"$ref": "#/$defs/path_absolute"},
+                                "url": {
+                                    "type": "string",
+                                    "pattern": "^((re:.*)?|(__[A-Z_]*__)?\/.*)$"
```
OniriCorpe, line 418: edit it here, sorry for my auto-formater

Psycojoker (approved):

OniriCorpe:
todo: same but for `additional_urls`

PR #2174, opened 2024-03-24
Move garradin to graveyard

Salamandar:
Repo was archived, and it's replaced by another app.

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -1285,17 +1285,6 @@ state = "working"
 subtags = [ "backup" ]
 url = "https://github.com/YunoHost-Apps/garage_ynh"
 
-[garradin]
-added_date = 1674232499 # 2023/01/20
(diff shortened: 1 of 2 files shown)


apps.toml
```diff
@@ -1275,17 +1275,6 @@ state = "working"
 subtags = [ "backup" ]
 url = "https://github.com/YunoHost-Apps/garage_ynh"
 
-[garradin]
-added_date = 1674232499 # 2023/01/20
-antifeatures = [ "deprecated-software", "replaced-by-another-app" ]
```
OniriCorpe, line 1280: maybe still keep it as deprecated for one or two months?

Tagadda: eeeh we could even have a special message in the diagnostic for apps in the graveyard

PR #2194, opened 2024-03-29
Add immich to catalog

ewilly:
immich is ready

@alexAubin : l'appli necessite postgresql16 et pour cela je dois installer des paquets depuis le repo debian de postgresql et donc mettre à jour des paquets issus de debian par ex : `postgresql-common` (cf.` [manifest](https://github.com/YunoHost-Apps/immich_ynh/blob/master/manifest.toml)). A prior je ne casse rien par ailleurs vis-à-vis des autres apps mais un avis d'expert me conforterais :-) peux-tu y jeter un coup d'oeil ?

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -1638,6 +1638,11 @@ state = "working"
 subtags = [ "accounting" ]
 url = "https://github.com/YunoHost-Apps/ihatemoney_ynh"
 
+[immich]
+category = "multimedia"
+state = "working"
+url = "https://github.com/YunoHost-Apps/immich_ynh"
+
 [incus]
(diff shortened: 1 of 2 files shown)


Thovi98:
Thanks for your work on this package!
If I understand well, you install also a new python version?

And the license in the manifest seems to be wrong, as Immich has AGPL-3.0 and not MIT :)

Salamandar:
TBH i don't really care about the internals of the package to merge this MR. I say let's go, and i will see on the package's merge requests how it's done.

Salamandar (approved):

PR #2197, opened 2024-03-30
Regen readme testing branch

Psycojoker:
Picks the testing branch to push new README if it exists.

Maybe we want both master and testing branch?


Psycojoker:
Oniricorpe tells me that pushing on testing is the implicit standard for this situation so I'll probably end up merging this PR.

tituspijean (approved):

PR #2217, opened 2024-04-03
Yologen : format python code with Black

lapineige:


lapineige:
How can I remove the changed to apps and whishlist.toml ?

PR #2233, opened 2024-04-09
Regenerate existing READMEs even if translation status has dropped from a lower value than 100%

oleole39:
This is a proposal to improve current behavior which is not to regenerate the README of the given language if that language has not been fully translated on Weblate.
It leads to cases where full translation of a given language was available, but then did not get updated. In this case, a README remains in the repo, not updated and not linked anymore in ALL_README.md.
With this PR, the readme would get generated whatever the current translation status (if no error) but add a warning to the README if not fully translated.

- [ ] It has not been tested
- [ ] Will L.200 get translated or should it be added to the hack section in ALL_README.md.j2 ? 
- [ ] Is it a good idea to make DESCRIPTION and DISCLAIMER block impact translation_warning variable (L.127 & L.138), knowing it is not the same translation process (one via Weblate, the other via Github) ?


tools/readme_generator/make_readme.py
```diff
@@ -106,7 +106,7 @@ def generate_READMEs(app_path: Path):
                 continue
             screenshots.append(str(entry.relative_to(app_path)))
 
-    def generate_single_README(lang_suffix: str, lang: str):
+    def generate_single_README(lang_suffix: str, lang: str, translation_warning: bool = False):
```
Psycojoker, line 110: Why do you pass "translation_warning" as a function parameter? Reading the code it looks like this should simply be a variable in the function '-'

Psycojoker: Ok, I got it reading the rest of the code ^w^

Psycojoker:
I need to test it but LGTM for now :)

tools/readme_generator/make_readme.py
oleole39: Let me detail this for future reference; translation_warning is: 
- passed to `generate_single_README()` as parameter when the warning relates to the weblate translation (readme structure)
- defined within `generate_single_README()` when the warning relates to the github translation (description.md and disclaimer.md). Not sure though if you will want to keep this behavior - cf. the 3rd checkbox in my message hereinabove.

PR #2269, opened 2024-04-29
Fix log formatting for notifications

tituspijean:
yunohost-bot is sending strings with unreplaced "%s" over Matrix.

I took also the opportunity to add a distinction between the two scripts.

### DIFF
--- a/tools/list_builder.py
+++ b/tools/list_builder.py
@@ -65,7 +65,7 @@ def __build_app_dict(data) -> Optional[tuple[str, dict[str, Any]]]:
     try:
         return name, build_app_dict(name, info)
     except Exception as err:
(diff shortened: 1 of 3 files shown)


tools/list_builder.py
```diff
@@ -65,7 +65,7 @@ def __build_app_dict(data) -> Optional[tuple[str, dict[str, Any]]]:
     try:
         return name, build_app_dict(name, info)
     except Exception as err:
-        logging.error("Error while updating %s: %s", name, err)
+        logging.error("[List builder] Error while updating %s: %s)", name, err)
```
OniriCorpe, line 68: is the added parenthesis intentionnal?

tituspijean: oops no

PR #2294, opened 2024-05-03
Add link to hosted YoloGen

orhtej2:
Why was it missing from README in the first place:

https://appgenerator.yunohost.org/


OniriCorpe:
because it was initialy hosted on @lapineige own domain name

lapineige:
And not launched publicly *yet*.

PR #2335, opened 2024-05-13
Increase items limit for autoupdater

tituspijean:
Untested.

Maybe we could be smarter/leaner and switch to 100 if first request returned nothing?

### DIFF
--- a/tools/autoupdate_app_sources/rest_api.py
+++ b/tools/autoupdate_app_sources/rest_api.py
@@ -6,11 +6,11 @@ from typing import Any, Optional
 
 import requests
 
+ITEMS_LIMIT=100
 
 class RefType(Enum):
     tags = 1
     commits = 2
-
 
 class GithubAPI:
     def __init__(self, upstream: str, auth: Optional[tuple[str, str]] = None):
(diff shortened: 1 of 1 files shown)


tools/autoupdate_app_sources/rest_api.py
```diff
@@ -29,15 +29,15 @@ def internal_api(self, uri: str) -> Any:
 
     def tags(self) -> list[dict[str, str]]:
         """Get a list of tags for project."""
-        return self.internal_api(f"repos/{self.upstream_repo}/tags")
+        return self.internal_api(f"repos/{self.upstream_repo}/tags?per_page="+str(ITEMS_LIMIT))
```
Salamandar, line 32: Just use the already present f-string ;)

Salamandar:
Just to understand. What's the "use case" ? If a newer version is released, it will most probably be in the first few tags (sorted by newest first).

tituspijean:
(Had to dig in the Matrix room history: https://matrix.to/#/!PauySEslPVuJCJCwlZ:matrix.org/$Vm4Leoc-esOggY90EbrQ5SsKfBM8ljEDS6gOv3apPPw?via=matrix.org&via=aria-net.org)

Use case is for apps releasing dozens of beta versions, effectively hiding the ones we want. In this case, bazarr.

PR #2338, opened 2024-05-14
Make RSS feed valid

oleole39:
Hello,

Although untested, this PR aims at fixing the issue reported there : https://github.com/alexAubin/tartiflettev2/issues/10#issuecomment-2105979218

@alexAubin

### DIFF
--- a/store/templates/news_rss.xml
+++ b/store/templates/news_rss.xml
@@ -9,17 +9,17 @@
 >
 	<channel>
 		<title>YunoHost apps catalog News</title>
-		<link>{{ url_for('news_rss')}}</link>
-		<atom:link href="{{ url_for('news_rss')}}" rel="self" type="application/rss+xml" />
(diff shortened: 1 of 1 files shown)


oleole39:
Hi @alexAubin 
Thank you for merging.
I know this is all beta, but FYI in prod https://apps.yunohost.org/news.rss now returns an "Internal Servor Error", without more details.
Last time I tried accessing it before this merge it returned the expected XML file (not a valid RSS structure, but an existing one). Not sure whether this can be due to this particular merge... Would `url_for ('news_rss', external=True)` (expected to return the absolute URI instead of a relative one) cause the error ? I did not tested it although I found some samples using it and that it is documented in Flask's docs...

alexAubin:
Indeed, thanks for the report, that's fixed via https://github.com/YunoHost/apps/pull/2343 :+1:

PR #2378, opened 2024-06-01
webhook: rebase testing on master if fast-forwardable

Salamandar:
# The issue

When merging testing on master on apps, we are left with testing "out of date" compared to master. So the testing branch lives parallel to master.

# The suggestion

Fast-forward the testing branch to master when it's possible. 

That means : if testing is an absolute parent commit of master, testing will then point to the same commit as master.
That does *not* mean : rebasing a testing branch where there is work not merged on master (and thus breaking the dev's workflow).

I splitted the PR in readable commits, i think the main PR diff is garbage :D


tituspijean (approved):
LGTM, and much needed.

PR #2399, opened 2024-06-09
Remove microblogpub wishlist

chri2:
added to catalog


chri2:
had been on the wishlist (didn't know)

PR #2404, opened 2024-06-10
Set Piwigo as not working

ericgaspar:
- Due to many complaints, set Piwigo as not working until it is fixed.


Salamandar:
crap. When i have time i can try to understand what's wrong.

OniriCorpe (approved):
fine by me

PR #2428, opened 2024-06-21
add autopatch: do not remove logs on app removal

OniriCorpe:


alexAubin (approved):
Naively LGTM but the real issue is "launching the autopatch" (i mean we can merge but will somebody run it to create the PRs ? ;P) 

i didnt do it in ages, I think last time i tried the python code wasnt reliable, but it was 2 years ago, dunno if somebody improved it

PR #2472, opened 2024-07-14
Move LXD to graveyard

Salamandar:
Considering we now have incus and lxd is kinda hard to maintain because we need to build it ourselves


kay0u:
Idk, I think I'll maintain it for a while.

Salamandar:
Alright, fine with me :)

PR #2548, opened 2024-08-30
Automatic patch/PR for helpers 2.1

alexAubin:


tools/autopatches/autopatch.py
```diff
@@ -18,16 +17,27 @@
...
 
-login = (
-    (REPO_APPS_ROOT / "tools/.github_login").open("r", encoding="utf-8").read().strip()
-)
-token = (
-    (REPO_APPS_ROOT / "tools/.github_token").open("r", encoding="utf-8").read().strip()
-)
+try:
+    login = (
+        (REPO_APPS_ROOT / "tools/.github_login").open("r", encoding="utf-8").read().strip()
```
jedie, line 33: It's a Path() instance, isn't it?
So:
`(REPO_APPS_ROOT / "tools/.github_login").read_text().strip()`
see: https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

Salamandar: Oh god that's an amazing api !

PR #2553, opened 2024-09-02
Add logos_check.py to validate logos filesize/resolution

Salamandar:
See https://github.com/YunoHost/issues/issues/2438

### DIFF
--- a/tools/logos_check.py
+++ b/tools/logos_check.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+
+import sys
+import argparse
+from pathlib import Path
+
+from PIL import Image
+
+
+class ImageCheck():
+    def __init__(self, imgpath: Path) -> None:
+        self.imgpath = imgpath
+        self.image = Image.open(imgpath)
+
+        self.fails: list[str] = []
+
+    def check(self) -> bool:
+        self.fails.clear()
+        for fn_name in dir(self):
+            if fn_name.startswith("check_"):
+                getattr(self, fn_name)()
+
+        if not self.fails:
+            return True
+
+        print(f"Image '{self.imgpath}' failed tests:")
+        for fail in self.fails:
+            print(f"  - {fail}")
+        print()
+        return False
+
+    def check_type(self) -> None:
+        format = self.image.format.lower()
+        accepted_formats = ["png", "jpeg", "webp"]
+        if format not in accepted_formats:
+            self.fails.append(f"Image should be one of {', '.join(accepted_formats)} but is {format}")
+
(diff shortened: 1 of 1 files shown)


tools/logos_check.py
```diff
@@ -0,0 +1,75 @@
...
+class ImageCheck():
+    def __init__(self, imgpath: Path) -> None:
+        self.imgpath = imgpath
+        self.image = Image.open(imgpath)
+
+        self.fails: list[str] = []
+
+    def check(self) -> bool:
+        self.fails.clear()
+        for fn_name in dir(self):
```
selfhoster1312, line 19: Explicitly listing the steps is clearer :)

tools/logos_check.py
```diff
@@ -0,0 +1,75 @@
...
+
+        print(f"Image '{self.imgpath}' failed tests:")
+        for fail in self.fails:
+            print(f"  - {fail}")
+        print()
+        return False
+
+    def check_type(self) -> None:
+        format = self.image.format.lower()
+        accepted_formats = ["png", "jpeg", "webp"]
```
selfhoster1312, line 34: Should we accept all those formats? Many software don't support web yet. And JPG doesn't allow background transparency which is nice for dark/light background. I think only PNG should be supported, but we could discuss it on the issue.

tools/logos_check.py
```diff
@@ -0,0 +1,75 @@
...
+        if format not in accepted_formats:
+            self.fails.append(f"Image should be one of {', '.join(accepted_formats)} but is {format}")
+
+    def check_dimensions(self) -> None:
+        dim_min, dim_max = 96, 300
```
selfhoster1312, line 39: Could we maybe hardcode those constants at the top of the file for easier changes?

tools/logos_check.py
```diff
@@ -0,0 +1,75 @@
...
+            self.fails.append(f"Dimensions should be in [{dim_min}, {dim_max}] but are {w}×{h}")
+
+    def check_ratio(self) -> None:
+        w, h = self.image.width, self.image.height
+        if w != h:
+            self.fails.append(f"Image is not square but {w}×{h}")
+
+    def check_filesize(self) -> None:
+        filesize = self.imgpath.stat().st_size
+        max_size = 80_000
```
selfhoster1312, line 52: Could we maybe hardcode this constant at the top of the file for easier changes?

tools/logos_check.py
alexAubin: (currently we have `.png` exclusively in `logos/` indeed)

PR #2624, opened 2024-10-14
Add Commet to catalog

ericgaspar:


tituspijean (approved):

apps.toml
```diff
@@ -542,6 +542,13 @@ state = "working"
 subtags = [ "website" ]
 url = "https://github.com/YunoHost-Apps/commento_ynh"
 
+[commet]
+category = "communication"
+potential_alternative_to = [ "Discord", "Facebook Messenger", "Signal", "Skype", "Telegram", "Whatsapp" ]
```
tituspijean, line 547:
```suggestion
potential_alternative_to = [ "Discord", "Element", "Cinny", "Facebook Messenger", "Signal", "Skype", "Telegram", "Whatsapp" ]
```

I do not know if we should add F(L)OSS apps to that list.

PR #2662, opened 2024-10-26
Lychee introduced paid features

ericgaspar:


OniriCorpe:
Uhhh should be `paid-content`, isn’t it?

PR #2684, opened 2024-11-06
Added Mitra

dragondaddy:
Added Mitra, another fediverse server using the ActivityPub protocol.


apps.toml
```diff
@@ -2584,6 +2584,13 @@ level = 6
 state = "working"
 url = "https://github.com/YunoHost-Apps/misskey_ynh"
 
+[mitra]
+category = "social media"
+potential_alternative_to = [ "Twitter", "Threads", "Bluesky" ]
+state = "inprogress"
```
alexAubin, line 2590: Note that "inprogress" means that it will be ignored by our CI ... imho it's an old flag which doesn't really make sense to use nowadays, either mark it as "working" (even if it's not entirely fully tested / working, but at least the CI will run - assuming you or somebody else intend to debug the issue) or "notworking"

dragondaddy: Right, "working" is probably the most appropriate choice then.

ericgaspar: I like your optimism, let's see how test goes

dragondaddy:
Hmm. What's wrong exactly here?..

ericgaspar (approved):

PR #2697, opened 2024-11-11
Add a subcategory for YunoHost apps and a bunch of apps

Salamandar:


alexAubin:
Hmmmokay but do we really want people to see those when browsing the catalog x_x

Salamandar:
I mean why not ? If they are in their own category, i don't think that causes any issue. Also that is a good way to show "we're open, you can recreate our infra if needed".

TBH the main reason I do that is to be able to do app updates without having to pass "--url" every time :D

alexAubin:
Because I can expect very much people randomly installing them and then asking support about it ...

OniriCorpe:
Maybe it’s time to support multiple sources catalog? And make a parallel catalog for our internal apps?

zeroheure:
Or add a "developer" group account, so that a group member could be able to see and install those apps.

lapineige:
Could they be a third option in the app list, to shown something like "YunoDev" app list ?

zamentur:
I close this PR cause YunoHost catalog is for end users. Currently we can't suggest to install an app like yunohost_myip_ynh (with private and public ip hardcoded). For some other packages we are not sure they work in all context for now.

However, i agree it could be nice to have myip, dynette and yunohost_diagnosis in the catalog in order to decentraliezd those parts.

SO i close, but if someone package a myip working for all situation, a dynette really working (maybe with a webui like dnsmanager did) or the diagnosis, i think we could rediscute it.

PR #2699, opened 2024-11-12
add phanpy to catalog

xmgz:
I think it's ready for install in any YNH


xmgz:
ok, added **phanpy** to Catalog.

Please tell me if further action is expected from my side **before** the app is actually included in the catalog to anyone install.


thank you

apps.toml
```diff
@@ -3212,6 +3212,12 @@ state = "working"
 subtags = [ "db" ]
 url = "https://github.com/YunoHost-Apps/pgweb_ynh"
 
+[phanpy]
+added_date = 1731396797 # 202/11/12
```
OniriCorpe, line 3216:
```suggestion
```

OniriCorpe (changes requested):

OniriCorpe (approved):

PR #2740, opened 2024-12-08
add marl_ynh

oleole39:
Added Mastodon Archive Reader Lite (MARL) to catalog.
Note that the app has currently no logo (hence no logo was added to YNH logos folder).


apps.toml
```diff
@@ -2406,6 +2406,11 @@ state = "working"
 subtags = [ "task" ]
 url = "https://github.com/YunoHost-Apps/mantis_ynh"
 
+[marl]
+category = "small_utilities"
+state = "working"
+url = "https://github.com/YunoHost-Apps/mastodon_ynh"
```
tituspijean, line 2412: That is not correct 😋

tituspijean (changes requested):

ericgaspar (approved):

apps.toml
oleole39: oops thanks for the edit @ericgaspar

tituspijean (approved):

PR #2763, opened 2025-01-08
AppFlowy

yalh76:


apps.toml
```diff
@@ -164,6 +164,13 @@ potential_alternative_to = [ "Capitalism" ]
 state = "working"
 url = "https://github.com/YunoHost-Apps/anarchism_ynh"
 
+[appflowy]
+category = "publishing"
+potential_alternative_to = [ "Notion" ]
```
tituspijean, line 169: Let's add an anti-feature, since it relies on OpenAI.

tituspijean (changes requested):

apps.toml
```diff
@@ -164,6 +164,13 @@ potential_alternative_to = [ "Capitalism" ]
 state = "working"
 url = "https://github.com/YunoHost-Apps/anarchism_ynh"
 
+[appflowy]
+category = "publishing"
```
tituspijean, line 168:
```suggestion
antifeatures = [ "non-free-network" ]
category = "publishing"
```

tituspijean: Antifeature since it relies on OpenAI, even optionally.

tituspijean (approved):

PR #2802, opened 2025-02-05
Update esphome in apps.toml

WGrav01:
Switched to broken (for now) and fixed my spelling

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -1125,8 +1125,8 @@ url = "https://github.com/YunoHost-Apps/ergo_ynh"
 
 [esphome]
 added_date = 1738792904 # 2025/02/05
-catagory = "iot"
-state = "working"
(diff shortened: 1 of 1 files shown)


apps.toml
```diff
@@ -1125,8 +1125,8 @@ url = "https://github.com/YunoHost-Apps/ergo_ynh"
 
 [esphome]
 added_date = 1738792904 # 2025/02/05
-catagory = "iot"
-state = "working"
+category = "iot"
+state = "broken"
```
alexAubin, line 1129: hmmmm that's probably gonna trigger other issues somewhere :sweat_smile: The only 3 possible values are "working", "inprogress", "notworking"

WGrav01: K, will fix

PR #2817, opened 2025-02-11
Add missing app info

CodeShakingSheep:
Just spotted some apps in the wishlist which had missing info (e.g. added websites and YNH draft packages).


alexAubin:
Cheers !

PR #2884, opened 2025-03-22
Add Webhook to catalog

ericgaspar:


logos/webhook.png
OniriCorpe, line 1: do we want to put a white background?

PR #2936, opened 2025-04-25
move DumbWhois in utilities

zeroheure:
it is expected to be in utilities because you don't really use it for system managing


zeroheure:
You are fast!

PR #2952, opened 2025-04-26
move archive box to reading cat

zeroheure:
Like Wallabag. Better than utilities


apps.toml
```diff
@@ -4081,18 +4081,17 @@ url = "https://github.com/YunoHost-Apps/roundcube_ynh"
 
 [rspamd]
 added_date = 1730230504 # 2024/10/29
-category = "system_tools"
+category = "communication"
```
alexAubin, line 4084: eeeh sounds like there's a mixup with another previous commit / PR

PR #3009, opened 2025-06-04
Change cat or subtag to some apps in communication

zeroheure:
Add subtag to mlmmj, mlmmj-web and Prosody.
Move Question2answer and Gemserv to publishing category


alexAubin:
Cheers !

PR #3068, opened 2025-07-31
add sparkyfitness to catalog

Thovi98:
Before merging, could someone help me to categorise this license in the manifest?

https://github.com/CodeWithCJ/SparkyFitness/tree/main?tab=License-1-ov-file

Thanks!


tituspijean:
That license is indeed tricky. It looks like a MIT with a bit a GPL inspiration on the derivative work.
It definitely needs a "not totally free" tag for its  commercial use.

PR #3111, opened 2025-08-31
move Grocy to productivity and management category

zeroheure:
I am not sure of that, but my rationale is that Grocy is lost in the small utilities category and that it will be easier to find it in management. Additionaly, I added the business_and_ngos sub category.

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -1868,9 +1868,10 @@ url = "https://github.com/YunoHost-Apps/grist_ynh"
 [grocy]
 added_date = 1674232499 # 2023/01/20
 branch = "master"
-category = "small_utilities"
+category = "productivity_and_management"
 level = 8
 state = "working"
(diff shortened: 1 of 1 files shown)


apps.toml
```diff
@@ -1868,9 +1868,10 @@ url = "https://github.com/YunoHost-Apps/grist_ynh"
 [grocy]
 added_date = 1674232499 # 2023/01/20
 branch = "master"
-category = "small_utilities"
+category = "productivity_and_management"
 level = 8
 state = "working"
+subtags = [ "business_and_ngos" ]
```
alexAubin, line 1874: hmmm not sure about this one ? I think this subtag is more like ERP/business/acconting but this is a fridge manager ? :thinking:

OniriCorpe: i think we should better create a new "home organization" category with a "kitchen" subcategory

PR #3133, opened 2025-09-18
Update immich

ewilly:
Immich is not anymore flagged upstream as alpha.


Thovi98:
> Expect bugs and breaking changes.

It’s not alpha but I don’t think it’s already considered as stable? As the stable version doesn’t seem to be far away, I’d wait for it :)

ewilly:
Alpha vs stable could raise to a long discussion...  but i'm fine to stay that disclaimer :)

Thovi98 (approved):

PR #3142, opened 2025-09-23
Mark civicrm_drupal as broken

cmcaine:
See: https://github.com/YunoHost-Apps/civicrm_drupal_ynh/issues/13


OniriCorpe:
I tagged the issue with `linter warning`, so our bit will automatically flag it as broken
Thanks for the help

PR #3420, opened 2026-04-08
Rename synapse-admin to ketesa

Josue-T:
Related to https://github.com/YunoHost-Apps/synapse-admin_ynh/pull/107

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -2586,6 +2586,15 @@ state = "working"
 subtags = [ "password" ]
 url = "https://github.com/YunoHost-Apps/keeweb_ynh"
 
+[ketesa]
+added_date = 1674232499 # 2023/01/20
+branch = "master"
+category = "communication"
+level = 8
+state = "working"
+subtags = [ "chat" ]
+url = "https://github.com/YunoHost-Apps/ketesa_ynh"
+
 [khatru-pyramid]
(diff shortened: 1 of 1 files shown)


ericgaspar (approved):

Thatoo (approved):

orhtej2:
Is that needed? Minetest got rebranded to Luanti by replacing the (UI) strings https://github.com/YunoHost-Apps/minetest_ynh/pull/151

ericgaspar:
> Is that needed? Minetest got rebranded to Luanti by replacing the (UI) strings [YunoHost-Apps/minetest_ynh#151](https://github.com/YunoHost-Apps/minetest_ynh/pull/151)

I will say that Minetest hasn't been renamed properly yet.

Josue-T:
In my opinion I prefer doing a clean rename because:
- It's really easy to rename the app. The user just need to remove the old one an install the new one. There are no data in the app so no data to migrate.
- Keeping the old Id in the time could be quite confusing. In 3 year when people will try to find the data of the app and will expect to find a directory named ketesa and it will be synapse-admin this is not great
- Also, we need to keep in mind that ketesa is a fork of synapse-admin and synapse-admin is still a project even if it's less maintained. So it's like if we distribute nextcloud with the id owncloud. Currently it would not make sense.

PR #3429, opened 2026-04-13
Add mail_relay as in-progress app

rungeard:
Add the mail_relay app entry to apps.toml with state set to inprogress and add a matching catalog logo with a white background.


apps.toml
```diff
@@ -3018,6 +3018,14 @@ state = "working"
 subtags = [ "email" ]
 url = "https://github.com/YunoHost-Apps/mailman3_ynh"
 
+[mail_relay]
+added_date = 1776106885 # 2026/04/13
+category = "communication"
+level = 0
```
```suggestion
```

Level should not be set, it's updated automatically, other than that LGTM :)

apps.toml
```diff
@@ -3018,6 +3018,13 @@ state = "working"
 subtags = [ "email" ]
 url = "https://github.com/YunoHost-Apps/mailman3_ynh"
 
+[mail_relay]
+added_date = 1776106885 # 2026/04/13
```
```suggestion
```
Same for added_date

PR #3447, opened 2026-04-24
pureblog is not totally free (non commercial license)

Thovi98:
https://github.com/kevquirk/pureblog/?tab=License-1-ov-file

If you have an idea what to put sa license in the manifest.toml, I’m all ear :D


tituspijean:
Custom licenses identifiers can be defined, and start with `LicenseRef-`. I suggest `LicenseRef-Pure-Blog-1`.
cf. https://github.com/YunoHost/issues/issues/2348
However, this is not fully integrated yet, as the license URL will be broken, according to that issue.

Edit: I misread the situation, this is completely unsupported so far. I'm working on it now.

Thovi98:
Thanks a lot! In the meantime I’ll already merge this :)

PR #3482, opened 2026-05-10
remove docmost from catalog

Thovi98:
See https://github.com/YunoHost-Apps/docmost_ynh/pull/1#issuecomment-4156494988

It seems that to package docmost, we have to build the `/ee` folder which is not free.
https://github.com/docmost/docmost?tab=readme-ov-file#license


Thovi98:
orhtej2 solved this :D

orhtej2:
TBF I think it was fixed upstream between 0.80.2 and 0.90.1

PR #3494, opened 2026-05-23
add(security): PeerTube SQL injection vulnerabilty

tituspijean:


tituspijean:
The title does not explain the whole story, but I preferred to keep it simple. More information on the provided link to the release.

PR #3496, opened 2026-05-24
Update prosody metadata and base branch

pitchum:


apps.toml
```diff
@@ -4516,9 +4516,10 @@ url = "https://github.com/YunoHost-Apps/prose_ynh"
 
 [prosody]
 added_date = 1674232499 # 2023/01/20
-branch = "master"
+branch = "stable"
```
Thovi98, line 4519:
```
branch = "stable"
```
Hmm I'm not sure this is supported? AFAIK branches `master` and `main` are the only one supported, but I may be wrong

alexAubin: The neutrinet app has `branch = 'stable'` since a long time, that should work

Thovi98: Okay then :D

apps.toml
```diff
@@ -4516,9 +4516,10 @@ url = "https://github.com/YunoHost-Apps/prose_ynh"
...
 category = "communication"
 level = 8
+potential_alternative_to = [ "Discord", "Facebook Messenger", "Signal", "Skype", "Telegram", "Whatsapp" ]
```
```suggestion
potential_alternative_to = [ "Discord", "Facebook Messenger", "Signal", "Skype", "Telegram", "WhatsApp" ]
```

PR #3564, opened 2026-07-29
Seerr renaming

ericgaspar:


Thovi98:
Thank you! But it's already done https://github.com/YunoHost/apps/pull/3557

PR #3568, opened 2026-07-31
Add mtg-multi to the catalog

mrPsycho:
This package differs from mtg by support of users. 
each user in yunohost now can see his own key to connect to mtproxy

### DIFF
--- a/apps.toml
+++ b/apps.toml
@@ -3554,6 +3554,12 @@ state = "working"
 subtags = [ "proxy" ]
 url = "https://github.com/YunoHost-Apps/mtg_ynh"
 
+[mtg-multi]
+category = "small_utilities"
+state = "working"
+subtags = [ "proxy" ]
+url = "https://github.com/YunoHost-Apps/mtg-multi_ynh"
+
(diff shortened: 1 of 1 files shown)


ericgaspar:
Is mtg-multi_ynh feature not worth integrating to mtg_ynh? (I don't know the ins and outs of mtg)

mrPsycho:
as these are quite new packages for me, i would say no. Why? because mtg_multi is a fork of mtg, and it may evolve in different way. 
and atm i'm using both: simple mtg just for personal instance, and mtg_multi for family needs.
but question is valid, yes.

apps.toml
```diff
@@ -3554,6 +3554,12 @@ state = "working"
 subtags = [ "proxy" ]
 url = "https://github.com/YunoHost-Apps/mtg_ynh"
 
+[mtg-multi]
```
orhtej2, line 3557: you need to specify the default branch, `main` IIRC is not the default one out of the box

orhtej2, line 3557: Consider adding logo file

PR #3598, opened 2026-09-04
Add gopodder to the catalog

M-Muq:
goPodder is a lightweight implementation of the gpodder API.

Status: Functional, tested for personal use


ericgaspar:
!invite @M-Muq

PR #3618, opened 2026-09-20
Add LibreDB Studio to the catalog

yusuf-gundogdu:
LibreDB Studio is a web SQL client you run on your own server: browse a schema, read and edit rows in a grid, write queries with completion, read EXPLAIN plans, draw an ER diagram. It speaks 16 engines through their own drivers, including PostgreSQL, MySQL and MariaDB, SQLite and libSQL, SQL Server, Oracle, MongoDB, Redis, ClickHouse and DuckDB.

The catalog has adminer, adminerevo, pgadmin, phpmyadmin and metabase under `system_tools` / `db`. This one sits beside them and covers the engines they do not.

Package: https://github.com/libredb/libredb_studio_ynh

## What I tested

A YunoHost 12.1.41 instance, amd64 packaging installed on arm64. Install, upgrade, backup, remove and restore, each from the repository URL rather than a local copy, all exit 0. After restoring I deleted the data directory first so that the restore had to actually put it back rather than find it already there.

The application itself, not just the service: signed in over the domain, listed connections, read the schema of the sample database, ran a join with a GROUP BY, and checked that a connection saved before an upgrade is still there after it.

`package_linter` passes every section. The one remaining hint is the systemd hardening one, which `example_ynh` triggers too: the check greps for `^\s<directive>=`, and no real unit indents its directives.

## Notes for a reviewer

**Domain root only, no path argument.** The payload is the upstream standalone tarball, and its route prefix is fixed when the image is built, so under a sub-path every asset 404s. I checked this in the shipped `server.js` rather than assuming it.

**Node 24.** `[resources.nodejs]`, because `package.json` declares `engines.node >= 24`. Most apps here are on 18 to 22, so this is the part I expect a question about. `n` fetched 24.21.0 without complaint.

**No LDAP, no SSO, permission defaults to `visitors`.** The application authenticates its own users, with local passwords, optional OIDC and per-account TOTP. It never sees the SSO headers, and `proxy_params_no_auth` strips them.

**The first run generates the administrator password and prints it once.** It goes to the service log, which the unit keeps at 0640 inside a 0750 directory, and to a 0600 file in the data directory. `doc/POST_INSTALL.md` says where to look.

**The repository is under our own organisation rather than YunoHost-Apps.** Happy to transfer it if you would rather the community maintained it; say the word and I will start the handover.


apps.toml
```diff
@@ -2915,6 +2915,15 @@ level = 8
 state = "working"
 url = "https://github.com/YunoHost-Apps/librebooking_ynh"
 
+[libredb_studio]
+added_date = 1789932510 # 2026/09/20
```
ericgaspar: That is add automatically

ericgaspar:
!invite @yusuf-gundogdu

yusuf-gundogdu:
Removed, thanks. I had not realised save_added_date.py fills it in from the git history.

Thanks for the org invite as well. I will accept it and transfer the repository, then update the url here.

yusuf-gundogdu:
Transferred to YunoHost-Apps/libredb_studio_ynh and the url here now points at it. testing and master are at the same commit.

Happy to run the CI on it, just tell me where you would like the !testme, since master already carries everything and a testing to master PR would be empty.

PR #3619, opened 2026-09-20
Add MRBS to the catalog

felurx:
This is my first package, any feedback is welcome (but a quick merge is also fine of course :P)


apps.toml
```diff
@@ -3649,6 +3649,12 @@ state = "working"
 subtags = [ "microblogging" ]
 url = "https://github.com/YunoHost-Apps/movim_ynh"
 
+[mrbs]
+branch = main
```
```suggestion
branch = "main"
```

Thovi98:
Waiting for https://github.com/YunoHost-Apps/mrbs_ynh/pull/6#issuecomment-5755793818

apps.toml
felurx: Whoops, thanks for the catch!

